PR #14, opened 2021-05-19
Add ECR policies (PLAT-2132)

ddl-denis-parnovskiy:
Goals:
- We want to give access to the same deployment images.
- We definitely want some specific EKS images. This comes for free because they are public and IAM doesn't affect them.
- We do not want to give a blanket access to each and every domino images in the same account because there can be images that are not in a deployment but still restricted.
- We do not care to limit to the account, this probably happens automatically because of "private" thing, we just do not want access to all images in the same account unless explicitly allowed.

Change summary:
1. Add ECR creation policy to the deploying role (this is not used right now but might be useful later).
2. Add explicit ECR deny policy that will prohibit pulling images from the same account if they are not marked as same deployment.
3. Make NodeGroup role names shorter and more human-readable.

Tests (our deploy: denis-test2):
- 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/coredns:v1.8.0-eksbuild.1 - allow
- 890728157128.dkr.ecr.us-west-2.amazonaws.com/denis-test2/nucleus:latest tags [denis-test2] - allow
- 890728157128.dkr.ecr.us-west-2.amazonaws.com/denis-test2/nucleus:latest tags [] - deny
- 890728157128.dkr.ecr.us-west-2.amazonaws.com/denis-test2-internal:latest tags [denis-test2-internal] - deny

### DIFF
--- a/cdk/domino/domino_stack.py
+++ b/cdk/domino/domino_stack.py
@@ -11,6 +11,7 @@ from typing import Any, Dict, Optional, Tuple
 
 import aws_cdk.aws_backup as backup
 import aws_cdk.aws_ec2 as ec2
+import aws_cdk.aws_ecr as ecr
 import aws_cdk.aws_efs as efs
 import aws_cdk.aws_eks as eks
 import aws_cdk.aws_events as events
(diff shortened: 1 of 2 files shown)


cdk/domino/domino_stack.py
```diff
@@ -323,7 +342,26 @@ def provision_eks_cluster(self):
...
+                self.s3_policy,
+                self.autoscaler_policy,
+                iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKSWorkerNodePolicy'),
+                iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEC2ContainerRegistryReadOnly'),
+                iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKS_CNI_Policy'),
+            ]
+            if self.config["route53"]["zone_ids"]:
+                managed_policies.append(self.route53_policy)
+
+            ng_role = iam.Role(
```
Secretions, line 356: If we're going to define the role from scratch rather than let EKS create it, maybe it'd make sense to just make a single role and assign that to all the managed/unmanaged nodegroups?

Secretions (approved):
We should add the ecr pip module to the Pipfile/setup.py as well, but LGTM for the deny policy. How do we still need to handle push?

cdk/domino/domino_stack.py
ddl-denis-parnovskiy: I was thinking about it.

I leave it to the case when managed and unmanaged nodegroup actually has some differences, including the roles.

Perhaps one role for all managed and one role for all unmanaged. I'll do it as a separate change

So far I define a single policy and attack it to the both roles.

PR #19, opened 2021-06-02
Misc dev/access improvements

Secretions:
PLAT-2246/PLAT-2247/PLAT-2230...

* dev deploys (proper reduced sizing)
  * Because the control plane requires a minimum of 2 AZs, just reducing to 1 AZ like we've done in the past doesn't work. Need to _only_ scale down the ASG AZs instead.
* Add SSH keypair support
* Add SSM support
* Add an optional bastion instance
* Fix private API setting

### DIFF
--- a/cdk/domino_cdk/config_template.yaml
+++ b/cdk/domino_cdk/config_template.yaml
@@ -59,9 +59,11 @@ eks:
     #   instance_types:
     #     - m5.2xlarge
     #   labels:
-    #     dominodatalab.com/node-pool: platform
+    #     dominodatalab.com/node-pool: default
+    #     domino/build-node: "true"
     #   tags:
-    #     dominodatalab.com/node-pool: platform
+    #     dominodatalab.com/node-pool: default
+    #     domino/build-node: "true"
   nodegroups:
     platform:
       gpu: false
--- a/cdk/domino_cdk/config_template.yaml
+++ b/cdk/domino_cdk/config_template.yaml
@@ -43,7 +43,25 @@ eks:
   max_nodegroup_azs: 1
   global_node_labels:
     dominodatalab.com/domino-node: "true"
+  # We support managed nodegroups, but recommend standard ones
   managed_nodegroups: {}
+    # managed_nodegroup_example:
+    #   spot: false
+    #   disk_size: 20
+    #   # key_name: some-aws-keypair
+    #   min_size: 1
+    #   max_size: 1
+    #   desired_size: 1
+    #   # machine_image:
+    #   #   ami_id: ami-xxxxx
+    #   #   user_data: |
+    #   #     data goes here
+    #   instance_types:
+    #     - m5.2xlarge
+    #   labels:
+    #     dominodatalab.com/node-pool: platform
+    #   tags:
+    #     dominodatalab.com/node-pool: platform
   nodegroups:
     platform:
       gpu: false
--- a/cdk/domino_cdk/config_template.yaml
+++ b/cdk/domino_cdk/config_template.yaml
@@ -11,7 +11,22 @@ vpc:
   id: null # ignored if create=true
   create: true
   cidr: "10.0.0.0/16"
-  max_azs: 3
+  max_azs: 3 # Must be 2+, as EKS control plane can't provision in a single AZ
+  bastion:
+    enabled: true
+    #key_name: some-aws-keypair
+    instance_type: t2.micro
+    ingress_ports:
+      - name: "ssh"
+        from_port: 22
+        to_port: 22
+        protocol: "TCP"
+        ip_cidrs:
+          - "0.0.0.0/0"
+        protocol: "TCP"
+    #machine_image:
+    #  ami_id: xxx
+    #  user_data: null
 efs:
   removal_policy_destroy: true
   backup:
@@ -25,19 +40,22 @@ route53:
 eks:
   version: 1.19
   private_api: true
+  max_nodegroup_azs: 1
   global_node_labels:
     dominodatalab.com/domino-node: "true"
-  managed_nodegroups:
+  managed_nodegroups: {}
+  nodegroups:
     platform:
-      spot: false
-      disk_size: 20
+      gpu: false
+      disk_size: 100
+      # key_name: some-aws-keypair
       min_size: 1
-      max_size: 1
-      desired_size: 1
+      max_size: 10
       # machine_image:
       #   ami_id: ami-xxxxx
       #   user_data: |
-      #     data goes here
+      #     #!/bin/bash
+      #     echo ${ClusterName}
       instance_types:
         - m5.2xlarge
       labels:
@@ -45,11 +63,11 @@ eks:
       tags:
         dominodatalab.com/node-pool: platform
     compute:
-      spot: false
+      gpu: false
       disk_size: 100
-      min_size: 1 # min 0 not supported w/managed node groups
+      # key_name: some-aws-keypair
+      min_size: 0
       max_size: 10
-      desired_size: 1
       # machine_image:
       #   ami_id: ami-xxxxx
       #   user_data: |
@@ -63,10 +81,10 @@ eks:
       tags:
         dominodatalab.com/node-pool: default
         domino/build-node: "true"
-  nodegroups:
     nvidia:
       gpu: true
       disk_size: 100
+      # key_name: some-aws-keypair
       min_size: 0
       max_size: 10
       # machine_image:
@@ -79,7 +97,7 @@ eks:
       taints:
         nvidia.com/gpu: "true:NoSchedule"
       labels:
-        dominodatalab.com/node-pool": default-gpu
+        dominodatalab.com/node-pool: default-gpu
         nvidia.com/gpu: "true"
       tags:
         dominodatalab.com/node-pool: default-gpu
--- a/cdk/domino_cdk/config_template.yaml
+++ b/cdk/domino_cdk/config_template.yaml
@@ -47,6 +47,7 @@ eks:
   nodegroups:
     platform:
       gpu: false
+      ssm_agent: true
       disk_size: 100
       # key_name: some-aws-keypair
       min_size: 1
@@ -64,6 +65,7 @@ eks:
         dominodatalab.com/node-pool: platform
     compute:
       gpu: false
+      ssm_agent: true
       disk_size: 100
       # key_name: some-aws-keypair
       min_size: 0
@@ -83,6 +85,7 @@ eks:
         domino/build-node: "true"
     nvidia:
       gpu: true
+      ssm_agent: true
       disk_size: 100
       # key_name: some-aws-keypair
       min_size: 0
--- a/cdk/domino_cdk/config_template.yaml
+++ b/cdk/domino_cdk/config_template.yaml
@@ -47,7 +47,7 @@ eks:
   managed_nodegroups: {}
     # managed_nodegroup_example:
     #   spot: false
-    #   disk_size: 20
+    #   disk_size: 100
     #   # key_name: some-aws-keypair
     #   min_size: 1
     #   max_size: 1
--- a/cdk/domino_cdk/config_template.yaml
+++ b/cdk/domino_cdk/config_template.yaml
@@ -1,5 +1,6 @@
 name: domino
 #aws_region: "__FILL__"
+#aws_account_id: "__FILL__"
 availability_zones: []
 #  - "us-west-2a"
 #  - "us-west-2b"
--- a/cdk/domino_cdk/eks_stack.py
+++ b/cdk/domino_cdk/eks_stack.py
@@ -27,6 +27,8 @@ from requests import get as requests_get
 from yaml import dump as yaml_dump
 from yaml import safe_load as yaml_safe_load
 
+from domino_cdk.util import DominoCdkUtil
+
 manifests = [
     [
         "calico",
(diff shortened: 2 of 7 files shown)


cdk/domino_cdk/eks_stack.py
```diff
@@ -579,6 +580,8 @@ def provision_unmanaged_nodegroup(
                         "--register-with-taints={}".format(",".join(["{}={}".format(k, v) for k, v in taints.items()]))
                     )
                 options["bootstrap_options"] = eks.BootstrapOptions(kubelet_extra_args=" ".join(extra_args))
+            if cfg["ssm_agent"]:
```
ddl-denis-parnovskiy, line 583: Don't you want to move it before using user_data in LaunchTemplate to make the logic more obvious?

cdk/domino_cdk/eks_stack.py
```diff
@@ -318,61 +395,55 @@ def provision_eks_cluster(self):
...
-                tags=cfg["tags"],
-                node_role=self.ng_role,
-            )
-
-        for name, cfg in self.config["eks"]["nodegroups"].items():
-            cfg["labels"] = {**cfg["labels"], **self.config["eks"]["global_node_labels"]}
-            self.provision_unmanaged_nodegroup(name, cfg, eks_version)
+        ami_id, user_data = self._get_machine_image(name, cfg)
+        machine_image: Optional[ec2.IMachineImage] = None
+        if ami_id:
```
ddl-denis-parnovskiy, line 402: https://eksctl.io/usage/launch-template-support/

When a launch template is provided, the following fields are not supported: instanceType, ami, ssh.allow, ssh.sourceSecurityGroupIds, securityGroups, instancePrefix, instanceName, ebsOptimized, volumeEncrypted, volumeKmsKeyID, volumeIOPS, maxPodsPerNode, preBootstrapCommands, overrideBootstrapCommand and disableIMDSv1

I read it as: you cannot specify AMI if you use launch template with managed nodegroup.

ddl-denis-parnovskiy: This is probably a typo in the doc because https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html says 

The following fields can't be set in the API if you're using a custom AMI
* amiType
* releaseVersion
* version

cdk/domino_cdk/eks_stack.py
```diff
@@ -318,61 +395,55 @@ def provision_eks_cluster(self):
...
+                        ec2.BlockDevice(
+                            device_name="/dev/xvda",
+                            volume=ec2.BlockDeviceVolume.ebs(
+                                cfg["disk_size"],
+                                volume_type=ec2.EbsDeviceVolumeType.GP2,
+                            ),
+                        )
+                    ],
+                    machine_image=machine_image,
+                    user_data=ec2.UserData.custom(cdk.Fn.sub(user_data, {"ClusterName": self.cluster.cluster_name})),
```
ddl-denis-parnovskiy, line 423: Reading the docs I still get an impression that you can specify user_data in launch template without specifying a machine image.

E.g. https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html

Quote 1: You can supply Amazon EC2 user data in your launch template

Quote 2: If your organization needs to run a custom AMI due to specific security, compliance, or internal policy requirements, you can deploy such AMIs to managed node groups by using a launch template

I'll try to see if ami in launch template is optional

ddl-denis-parnovskiy: https://aws.amazon.com/blogs/containers/introducing-launch-template-and-custom-ami-support-in-amazon-eks-managed-node-groups/

it has a picture where AMI is selected as "Do not include in Launch Template" 

I still don't see how to do it :)

ddl-denis-parnovskiy: Yes, I can do it. 

AMI ID is optional

See: https://us-west-2.console.aws.amazon.com/ec2/home?region=us-west-2#LaunchTemplateDetails:launchTemplateId=lt-0aeeb44c90ed676d8

Launch template with user data but without AMI

cdk/domino_cdk/eks_stack.py
Secretions: I had the wrong indent on the original version of this you reviewed, but this is meant to only be executed when the code above it is executed.

cdk/domino_cdk/eks_stack.py
Secretions: I don't know that we should treat eksctl docs as representative of what's possible with either the eks, cloudformation or cdk apis.  This code is all code that was there before, so I believe Michael tested it with the other AMIs. I think it should be functional.

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: My point is: 
1. You pass a reference.
2. You update an object at the reference later.

This obscures the logic a little bit for the reader who is not aware if user_data is copied to the destination or is left as a reference.

I.e. instead of the code
```
envelope = Envelope()
mailbox.put(envelope)
envelope.letter = "hello"
```

I would write
```
envelope = Envelope()
envelope.letter = "hello"
mailbox.put(envelope)
```

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: I'll take this.

cdk/domino_cdk/eks_stack.py
```diff
@@ -494,6 +580,9 @@ def provision_unmanaged_nodegroup(self, name: str, cfg: dict, eks_version: eks.K
                         "--register-with-taints={}".format(",".join(["{}={}".format(k, v) for k, v in taints.items()]))
                     )
                 options["bootstrap_options"] = eks.BootstrapOptions(kubelet_extra_args=" ".join(extra_args))
+                
+                if cfg["ssm_agent"]:
+                    asg.user_data.add_on_exit_commands("yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm")
```
ddl-denis-parnovskiy, line 585: A modification of asg.user_data after it was "consumed" by LaunchTemplate at line 543 looks cheesy.

What is the effect and side effect?

Do you intend to modify asg AFTER its user data was used in LaunchTemplate?

Or do you pass an object by reference and do a backdated update of that object?

cdk/domino_cdk/eks_stack.py
```diff
@@ -318,61 +395,55 @@ def provision_eks_cluster(self):
...
+                    machine_image=machine_image,
+                    user_data=ec2.UserData.custom(cdk.Fn.sub(user_data, {"ClusterName": self.cluster.cluster_name})),
+                )
+                lts = eks.LaunchTemplateSpec(id=lt.launch_template_id, version=lt.version_number)
+                disk_size = None
+
+        ng = self.cluster.add_nodegroup_capacity(
+            f"{name}-{i}",  # this might be dangerous
+            nodegroup_name=f"{self.name}-{name}-{az}",  # this might be dangerous
+            capacity_type=eks.CapacityType.SPOT if cfg["spot"] else eks.CapacityType.ON_DEMAND,
```
ddl-denis-parnovskiy, line 431: this raises if 'spot' is not specified

use
```
cfg.get('spot')
```

cdk/domino_cdk/eks_stack.py
```diff
@@ -318,61 +395,55 @@ def provision_eks_cluster(self):
...
+            disk_size=disk_size,
+            min_size=cfg["min_size"],
+            max_size=cfg["max_size"],
+            desired_size=cfg["desired_size"],
```
ddl-denis-parnovskiy, line 435: same

cdk/domino_cdk/eks_stack.py
```diff
@@ -318,61 +395,55 @@ def provision_eks_cluster(self):
...
+            subnets=ec2.SubnetSelection(
+                subnet_group_name=self.private_subnet_name,
+                availability_zones=[az],
+            ),
+            instance_types=[ec2.InstanceType(it) for it in cfg["instance_types"]],
+            launch_template_spec=lts,
+            labels=cfg["labels"],
+            tags=cfg["tags"],
+            node_role=ng_role,
```
ddl-denis-parnovskiy, line 444: self.ng_role ?

cdk/domino_cdk/eks_stack.py
```diff
@@ -318,61 +395,55 @@ def provision_eks_cluster(self):
...
+            remote_access=eks.NodegroupRemoteAccess(cfg["key_name"]),
```
ddl-denis-parnovskiy, line 445: This is commented out in the template and I have no idea where to get it. This is not in README

cdk/domino_cdk/config_template.yaml
```diff
@@ -24,31 +40,36 @@ route53:
...
       tags:
         dominodatalab.com/node-pool: platform
     compute:
-      spot: false
+      gpu: false
+      ssm_agent: true
       disk_size: 100
-      min_size: 1 # min 0 not supported w/managed node groups
+      # key_name: some-aws-keypair
+      min_size: 0
```
ddl-denis-parnovskiy, line 71: 12:13:38 AM | CREATE_FAILED        | AWS::EKS::Nodegroup                   | eks/Nodegroupcompute-0
minSize can't be zero or negative

Secretions: I don't think your config is actually based on this template. You'll note I moved all the default nodegroups under the unmanaged nodegroups, but your error suggests they're in the managed ones.

cdk/domino_cdk/eks_stack.py
Secretions: I guess I should add some form of template back to the managed nodegroups section in the config, but spot should be in the template such that it's always specified (this is how it was before--this is _NOT_ new code).

cdk/domino_cdk/eks_stack.py
Secretions: I have a change I haven't pushed yet to make this key optional, but I presumed that the placeholder value in the template would make it clear enough for now (`some-aws-keypair`) that it's an aws keypair that's already provisioned in AWS. I can either add a more explicit comment in the template or add to the readme if you think it's needed.

cdk/domino_cdk/config_template.yaml
ddl-denis-parnovskiy: Correct

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: Agree.

Either add managed template example (commented out if you want), or use get with a healthy defaults.

I prefer defaults.

cdk/domino_cdk/eks_stack.py
Secretions: About how it's modifying it: `asg.user_data` is an attribute of the asg that's meant to be interacted with in this fashion, and adding via the add_on_exit_commands ensure that we are adding this after the main user_data script has been ran (and that we're not interfering with it).

About why we use the attribute: We do _not_ necessarily specify `user_data`, even though there is still a `user_data` script. If `user_data` is specified manually in the config, we'll use that. Otherwise, we'll let the default `user_data` in the ami be used, in which case we don't have access to the `user_data` until after the asg is created. You'll also note that `asg.user_data` is what's input into the launch template's user_data attribute.

cdk/domino_cdk/eks_stack.py
Secretions: I clearly need to test managed nodegroups again after the refactor. :p

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: Yes, I note all of this.

Let's try again. 

You do
```
if user_data:
  create_launch_template( user_data = asg.user_data )

if ssm:
  asg.user_data.append(stuff)
``` 

My grievance is: the effect of this sequence is unclear and largely depends on the internal logic of "create_launch_template".

If it copies user data, then effect of changing it in asg will affect asg only.

If it keeps user data as a reference, then effect of changing it in asg will have side effect of changing the launch template as well.

I suggest to make this side effect clear if it exists and desired.

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: If you make it optional, then all good.

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: Also

user_data will go to both LaunchConfiguration and LaunchTemplate. 

I cannot find distinctive answer in the doc but it looks like LaunchTemplate will override LaunchConfiguration

If user_data is copied, then your SSM code will be lost. 

I'll probably check this today

ddl-denis-parnovskiy: Correction :)

You do:
```
if user_data:
  create_launch_template( user_data = asg.user_data )
else:
  if ssm:
    asg.user_data.append(stuff)
```

I.e. ssm stuff is skipped if user_data is present.

I could not comprehend that this is kinda desired logic. Is it?

ddl-denis-parnovskiy: Let's move on.

I will probably change this part anyway resulting from my user_data investigation

Secretions: The launch template _always_ needs to be created, as this is the unmanaged nodegroup code, and we always delete the launch config and replace it with our launch template.

So full logic is this:

```
custom_ami_id, custom_user_data = get_custom_machine_image_settings()
machine_image = get_custom_machine_image_or_return_default_eks_machine_image()
asg = create_asg(machine_image)
lt = create_launch_template()
hack_to_use_launch_template_instead_of_launch_config(asg, lt)
if not custom_user_data: # we're using standard eks user data, so safe to do normal domino customizations
  options = setup_eks_bootstrap_args()
  if cfg["ssm_agent"]:
    asg.user_data.add_on_exit_commands("...")
self.cluster.connect_auto_scaling_group_capacity(asg, options)
```

Ignoring whether you might be able to move the `if not custom_user_data` block up a few lines, _this logic should not change_, at least WRT current functionality.

Secretions: I can update the old code to try and make all of this more clear.

ddl-denis-parnovskiy: ok, let's see

```
custom_ami_id, custom_user_data = get_custom_machine_image_settings()
if custom_user_data:
  machine_image = custom_machine_image( custom_user_data )
asg = create_asg(machine_image) # assuming asg.user_data is populated
lt = create_launch_template( user_data = asg.user_data )
hack_to_use_launch_template_instead_of_launch_config(asg, lt)
if not custom_user_data: # we're using standard eks user data, so safe to do normal domino customizations
  options = setup_eks_bootstrap_args()
  if cfg["ssm_agent"]:
    asg.user_data.add_on_exit_commands("...")
self.cluster.connect_auto_scaling_group_capacity(asg, options)
```
Here we make several not-so-obvious assumptions.

1. user_data from AMI will go into asg and will be immediately available as asg.user_data. I'm glad if this works. 

2. After creating the launch_template, we modify user_data in asg. But, launch_template, if specified, replaces launch_config. So my expectations here: after we have created launch_template, settings in asg.user_data are ignored. launch_template has no comment "user data is still mutable after creations" like user_data in asg.

I probably do not understand why 
1. if we use auto-assigned image, we are ok with user_data coming with the image
2. if we assign a specific image, we override user_data

Secretions: Assumption section:

1. user_data will get determined when creating the machine_image object, either it uses the user_data you provide it or it defaults to what's in the AMI. Since we pass the machine image object to the asg, that's where it gets it from.
2. I can change asg.user_data to lt.user_data if you prefer. They both do the same thing (I've tested both ways). I can also/alternatively move the entire if block (ie `if not user_data`) above the launch template definition, but I do think it's much nicer having the bootstrap info at the end. Would one of these solutions resolve this concern?

Understand section:

1. The image we're retrieving if an AMI wasn't specified is the default eks AMI, which has user_data specifically designed to bring up an EKS cluster. We need want to mess with it, it's meant to be used as-is, as if we manually specify user_data that means we now have to track the user_data of the official eks ami in case it changes.
2. If you assign a specific image, the assumption is that you're doing something like trying to get EKS to run off a stock RedHat AMI, in which case it's just flat out not going to work unless you setup user_data. We aren't trying to cover the case that someone generates an EKS-focused ami with the user_data setup. It's both extremely unlikely to ever come up, and muddies the waters for the officially supported thing vs the custom thing (and has an easy fix of "put the user_data in the config"--and the concern about user_data changing doesn't really matter here since it's specific to the specified ami).

ddl-denis-parnovskiy: Probably just a comment that user data in LT is still mutable at this point.

ddl-denis-parnovskiy: Thanks for the explanations about EKS vs random image

ddl-denis-parnovskiy: I still do not like that ssm will not work if the user data is custom. 

I understand that we expect that whoever provides custom_user data should write the whole deal, but perhaps SSM is special.

I understand that is it tricky to add SSM to a random bash script, but EKS deals with it with Mime, merging several user_data.

I still not not exactly understand how MIME works, though

Secretions: I switched it to `lt.user_data`, and added a comment about it. I also added an error condition so if users turn on things that would have previously been ignored with custom user_data, now they'll be notified.

We really can't even assume that a machine has yum, necessarily. I think it's on the wrong side of a philosophical line to be doing any implicit/hidden user_data manipulation under the scenes in the circumstance that the user is providing it. Hopefully the error condition making things explicit is a bit better than nothing, though.

cdk/domino_cdk/eks_stack.py
Secretions: Since we're not really going to use managed nodegroups anymore I wouldn't want them uncommented, but I added a commented example back in.

cdk/domino_cdk/eks_stack.py
Secretions: Tested managed nodegroups

cdk/domino_cdk/eks_stack.py
```diff
@@ -318,61 +400,56 @@ def provision_eks_cluster(self):
...
+                    machine_image=machine_image,
+                    user_data=ec2.UserData.custom(cdk.Fn.sub(user_data, {"ClusterName": self.cluster.cluster_name})),
+                )
+                lts = eks.LaunchTemplateSpec(id=lt.launch_template_id, version=lt.version_number)
+                disk_size = None
+
+        key_name = cfg.get("key_name", None)
+        self.cluster.add_nodegroup_capacity(
+            f"{name}-{i}",  # this might be dangerous
+            nodegroup_name=f"{self.name}-{name}-{az}",  # this might be dangerous
```
ddl-denis-parnovskiy, line 436: Is it my eyes or this construct using i and az is not inside of `for i, az` loop?

Secretions: Nope, not your eyes. Good catch, the indent somehow got off when I moved this into a function. While I did ultimately test the managed nodegroups, I did it with the nodegroup AZs set to one, so it serendipitously did the right thing.

ddl-denis-parnovskiy (approved):
No further comments

PR #21, opened 2021-06-04
increase platform node pool disk size to 100Gb

steved:
20Gb can quickly lead to contention from the number of images that make
up a Domino (especially when upgrading / changing images).


Secretions (approved):

Secretions:
Actually, realized I ended up merging these changes into my now-merged PR, so closing.

PR #22, opened 2021-06-09
Object-based config

Secretions:
* Adds an object (dataclass) based in-memory config, replacing the dict we blindly load. 
* Performs type checking when loading configs 
  * Though it won't enforce the nested types specified for lists/dicts
  * Type checks do produce errors
  * Errors display which config entry caused the error
* Config schemas are now versioned
  * Each config object has its own loader, versioned to the schema
  * As things change, we can re-map variables either in-schema or in-object as we need to and still be compatible with existing/old configs
  * In some long run, we may find we want to deprecate older schemas so we don't have to manage too many old loaders. We can feel for that as we go
  * (follow-up todo): A config file can be loaded, and then spat back out with the current schema
    * This gives a way to get to the current schema even if a schema version gets deprecated in the newest version (ie pull an older version that still supported the schema, upgrade it, then use current version).
  * Does not error out on extra variables in the config, but it will print warnings
* We can now generate the config file based on a function (similar to init in the agent)
  * Assumes production defaults, but there's a dev option
  * I have some thoughts on how to make something nicer to come up with a config from scratch, but that's for another day
  * Generated configs have documentation for every variable (this required using `ruamel.yaml`)
* The config functions now have unit tests
  * Reorganize the tests that are there right now
  * Individually test every schema version loader for every object, this will help keep us on the ball for compatibility with older versions

Stuff that should be done, but will be either follow up or for an other day entirely:

Follow-up:

* Add argparse to app.py instead of the janky sys.argv (or move non-cdk functions into additional utility program)
* Make all currently existing options in api for config generator/renderer available via cli
* Add version of generator/renderer that loads an existing config to facilitate schema upgrades
  * Should display some warning if schema is out of date with instructions on how to update
  * Don't update it in place automatically on behalf of the user, though. I think?
* Remove config template in favor of config generator

Another day:

* Automatically run unit tests via CI
* Full unit test coverage of the whole domino_cdk app
* Flexible config generator

image


cdk/domino_cdk/config/base.py
```diff
@@ -0,0 +1,140 @@
...
+            raise ValueError("\n".join(errors))
+
+    def render(self, disable_comments: bool = False):
+        def r_vars(c, indent: int):
+            indent += 2
+            if is_dataclass(c):
+                cm = CommentedMap({x: r_vars(y, indent) for x, y in vars(c).items()})
+                if not disable_comments:
+                    [
+                        cm.yaml_set_comment_before_after_key(k, after=v.__doc__, after_indent=indent)
```
steved, line 121: wrapping this in `dedent(v.__doc__).strip()` would allow you to specify the dataclass documentation just like regular python:
```
$ cat test.py
from textwrap import dedent

class ABC:
    class DEF:
        """
        this is my cool
        very indented doc
        """


print(ABC.DEF.__doc__)
print(dedent(ABC.DEF.__doc__).strip())

$ python test.py

        this is my cool
        very indented doc

this is my cool
very indented doc
```

cdk/app.py
```diff
@@ -40,9 +45,12 @@
             print(json_dumps(DominoCdkUtil.generate_asset_parameters(*argv[2:]), indent=4))
         elif argv[1] == "generate_terraform_bootstrap":
             print(json_dumps(DominoCdkUtil.generate_terraform_bootstrap(*argv[2:]), indent=4))
+        elif argv[1] == "generate_config_template":
```
steved, line 48: More a nitpick than anything, but at some point consider just adding argparse please?

cdk/app.py
```diff
@@ -1,37 +1,42 @@
...
-app = core.App()
 
-with open(app.node.try_get_context("config") or "config.yaml") as f:
-    y = f.read()
-    cfg = yaml.safe_load(y)
+def main():
+    app = core.App()
 
-env_vars = {}
+    with open(app.node.try_get_context("config") or "config.yaml") as f:
```
steved, line 17: this bails if config.yaml doesn't exist

cdk/domino_cdk/config/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+from domino_cdk import __version__
+from domino_cdk.config.base import DominoCDKConfig
+
+
+def config_loader(c: dict):
+    schema = Version(c.pop("schema", "0.0.0")).truncate()
+    if schema == Version("0.0.1"):
+        return DominoCDKConfig.from_0_0_1(c)
+    elif schema == Version("0.0.0"):
+        return DominoCDKConfig.from_0_0_0(c)
```
steved, line 12: would definitely prefer some sort of schema registration system so we don't need a lot of if/else (and could validate that the schema itself is a valid identifier)

cdk/domino_cdk/config/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+    platform_instance_type = "m5.2xlarge"
+    platform_min_size = 3
+    if dev_defaults:
+        max_nodegroup_azs = 1
+        destroy_on_destroy = True
+        disk_size = 100
+        platform_instance_type = "m5.4xlarge"
+        platform_min_size = 1
+    return config_loader(
+        {
```
steved, line 28: I realize it'll fail fast, but constructing objects for this inline makes me feel more comfortable (like the one in the unit test?)

cdk/domino_cdk/config/base.py
```diff
@@ -0,0 +1,140 @@
...
+                eks=EKS.from_0_0_1(c.pop("eks")),
+                s3=S3.from_0_0_0(c.pop("s3")),
+            ),
+            c,
+        )
+
+    def __post_init__(self):
+        errors = []
+
+        def val(path: str, obj):
```
steved, line 89: this is really cool :+1:

I wonder if generating a JSON schema or something similar from the dataclass itself could be a next step so we could defer the type checking validation earlier. On the other hand, this kind of construction would let us hook in custom validation through dataclass field `metadata` or something.

steved (commented):
Just a general thought, I wonder if https://github.com/konradhalas/dacite can somehow help with some of the boilerplate of loading the actual dict into the dataclass.

Would a v1 also include some way of dumping the transformed schema 0.0.0 to 0.0.1? What would our support model be for schema versioning?

cdk/app.py
steved: Just kidding, actually read your PR description.

cdk/app.py
Secretions: If you're doing cdk synth, it should bail if config.yaml doesn't exist. If you're doing one of the special args (ie getting the template), then it shouldn't be getting to this part of the code.

cdk/domino_cdk/config/base.py
Secretions: Good call

cdk/domino_cdk/config/__init__.py
Secretions: Yes, I'll be doing something with this in follow up.

cdk/domino_cdk/config/__init__.py
Secretions: It's less involved than what you were demonstrating in slack, but I made it dynamic and added some related schema version validation tests in the unit tests.

Secretions:
> Would a v1 also include some way of dumping the transformed schema 0.0.0 to 0.0.1? What would our support model be for schema versioning?

Yeah, it will be in a subsequent PR, but when I add argparse and do a better version of the config generation, it'll also including an option to load in the config and re-render it with the current schema.

ddl-denis-parnovskiy (approved):

cdk/domino_cdk/eks_stack.py
```diff
@@ -391,31 +405,24 @@ def provision_eks_iam_policies(self):
...
             managed_policies=managed_policies,
         )
 
-    def _get_machine_image(self, cfg_name: str, cfg: Dict[str, Any]) -> Tuple[Optional[str], Optional[str]]:
-        ami_id = cfg.get("ami_id")
-        user_data = cfg.get("user_data")
-        if ami_id and not user_data:
-            raise Exception("User data must be provided when specifying a custom AMI")
-        if ami_id and (cfg.get("ssm_agent") or cfg.get("labels") or cfg.get("taints")):
-            raise Exception(
```
Secretions, line 411: These exceptions were moved into the config object.

PR #23, opened 2021-06-11
Improvements for custom ami and user data handling (PLAT-2266 and PLAT-2354)

ddl-denis-parnovskiy:
If ami_id is not specified in managed user group, EKS automatically provides latest and greatest Linux2 image. AWS documentation vague hints that the nodes will be also automatically upgraded if a newer and greater Linux2 image emerges.

In this case EKS provides its bootstrap logic as a part in MIME/multipart and demands that any other user data was also in MIME/multipart so it can join the parts together.

user data is fed to cloud_init and MIME/multipart is one of the formats where multiple scripts are passed as parts.

Benefits for managed node groups:
1. We can have true auto-upgrading fully managed autoscaling group.
2. We can add "install ssm" logic and "tweak QPS in kubelet" as separate bash scripts that do not intersect with any custom user data.

Also, we can improve unmanaged node groups:
1. We can add "install ssm" logic and "tweak QPS in kubelet" as separate bash scripts that do not intersect with any custom user data (compared to the old approach that injects custom user data as a bash function in front of the standard script).
2. We can unify user data handling for managed and unmanaged groups.

BTW. For unmanaged groups. 
* there is no default user data script in default AMI. 
* all bootstrap code is generated by CDK and is part of the generated CF template 
* It is not even EKS-provided.

### DIFF
--- a/cdk/domino_cdk/eks_stack.py
+++ b/cdk/domino_cdk/eks_stack.py
@@ -491,9 +491,10 @@ class DominoEksStack(cdk.Stack):
                 lts = None
 
             indexed_name = f"{self.name}-{name}-{az}"
+            indexed_id = f"{self.name}-{name}-{i}"
             key_name = cfg.get("key_name", None)
             self.cluster.add_nodegroup_capacity(
-                indexed_name,  # this might be dangerous
+                indexed_id,  # this might be dangerous
                 nodegroup_name=indexed_name,
                 capacity_type=eks.CapacityType.SPOT if cfg["spot"] else eks.CapacityType.ON_DEMAND,
                 min_size=cfg["min_size"],
@@ -546,9 +547,10 @@ class DominoEksStack(cdk.Stack):
         scope = cdk.Construct(self, f"UnmanagedNodeGroup{name}")
         for i, az in enumerate(self.vpc.availability_zones[:max_nodegroup_azs]):
             indexed_name = f"{self.name}-{name}-{az}"
+            indexed_id = f"{self.name}-{name}-{i}"
             asg = aws_autoscaling.AutoScalingGroup(
                 scope,
-                indexed_name,
+                indexed_id,
                 auto_scaling_group_name=indexed_name,
                 instance_type=ec2.InstanceType(cfg["instance_types"][0]),
                 machine_image=machine_image,
--- a/cdk/domino_cdk/eks_stack.py
+++ b/cdk/domino_cdk/eks_stack.py
@@ -417,7 +417,6 @@ class DominoEksStack(cdk.Stack):
     def _handle_user_data(
         self, name: str, custom_ami: bool, cfg: dict, user_data_list: List[Union[ec2.UserData, str]]
     ) -> Optional[ec2.UserData]:
-        proceed: bool = False
         mime_user_data = ec2.MultipartUserData()
 
         # If we are using default EKS image, tweak kubelet
@@ -430,16 +429,15 @@ class DominoEksStack(cdk.Stack):
                     )
                 ),
             )
-            proceed = True
-        if cfg.get("ssm_agent"):
-            mime_user_data.add_part(
-                ec2.MultipartBody.from_user_data(
-                    ec2.UserData.custom(
-                        "yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm",
-                    )
-                ),
-            )
-            proceed = True
+            # if not custom AMI, we can install ssm agent. If requested.
+            if cfg.get("ssm_agent"):
+                mime_user_data.add_part(
+                    ec2.MultipartBody.from_user_data(
+                        ec2.UserData.custom(
+                            "yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm",
+                        )
+                    ),
+                )
         for ud in user_data_list:
             if isinstance(ud, str):
                 mime_user_data.add_part(
@@ -458,10 +456,6 @@ class DominoEksStack(cdk.Stack):
                 )
             if isinstance(ud, ec2.UserData):
                 mime_user_data.add_part(ec2.MultipartBody.from_user_data(ud))
-            proceed = True
-
-        if not proceed:
-            return None
 
         return mime_user_data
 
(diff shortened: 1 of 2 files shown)


cdk/domino_cdk/eks_stack.py
```diff
@@ -298,11 +298,13 @@ def provision_eks_cluster(self):
         max_nodegroup_azs = self.config["eks"]["max_nodegroup_azs"]
 
         for name, cfg in self.config["eks"]["managed_nodegroups"].items():
-            cfg["labels"] = {**cfg["labels"], **self.config["eks"]["global_node_labels"]}
```
ddl-denis-parnovskiy, line 301: fix of pre-existing bug. You cannot have labels with custom ami_id but you also cannot remove labels because of this line.

cdk/domino_cdk/util.py
```diff
@@ -13,7 +13,7 @@ class ExternalCommandException(Exception):
     """Exception running spawned external commands"""
 
 
-class DominoCdkUtil():
```
ddl-denis-parnovskiy, line 16: this is "black", not me

cdk/domino_cdk/eks_stack.py
```diff
@@ -498,7 +550,6 @@ def provision_unmanaged_nodegroup(
 
         scope = cdk.Construct(self, f"UnmanagedNodeGroup{name}")
         for i, az in enumerate(self.vpc.availability_zones[:max_nodegroup_azs]):
-
             indexed_name = f"{self.name}-{name}-{i}"
```
Secretions, line 553: This the area I mentioned in my other comment where I think we could switch to az. The `f"ASG{i}"` below also seems kind of bad, as we're naming multiple ASGs something non-unique. Apparently doesn't cause problems because it's been working, but still... throwing the asg name in as well (the model used for managed) would be better for that I think.

cdk/domino_cdk/eks_stack.py
```diff
@@ -400,17 +402,78 @@ def provision_eks_iam_policies(self):
...
+            mime_user_data.add_part(
+                ec2.MultipartBody.from_user_data(
+                    ec2.UserData.custom(
+                        'KUBELET_CONFIG=/etc/kubernetes/kubelet/kubelet-config.json\n'
+                        'echo "$(jq \'.eventRecordQPS=0\' $KUBELET_CONFIG)" > $KUBELET_CONFIG'
+                    )
+                ),
+            )
+            proceed = True
+        if cfg.get("ssm_agent"):
```
Secretions, line 434: Since we only do this with a custom ami, I wonder if we should indent this once? It's not really _needed_ because of the exception elsewhere, but organizationally I think it makes sense, as reading the code it might be more obvious what the intent is.

cdk/domino_cdk/eks_stack.py
```diff
@@ -400,17 +402,78 @@ def provision_eks_iam_policies(self):
...
+        proceed: bool = False
+        mime_user_data = ec2.MultipartUserData()
+
+        # If we are using default EKS image, tweak kubelet
+        if not custom_ami:
+            mime_user_data.add_part(
+                ec2.MultipartBody.from_user_data(
+                    ec2.UserData.custom(
+                        'KUBELET_CONFIG=/etc/kubernetes/kubelet/kubelet-config.json\n'
+                        'echo "$(jq \'.eventRecordQPS=0\' $KUBELET_CONFIG)" > $KUBELET_CONFIG'
```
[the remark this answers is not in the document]
ddl-denis-parnovskiy: This happens before kubelet setup. Before bootstrap.sh.

Why. Because bootstrap.sh or whatever does start kubelet. But kubelet is installed in the image and kubelet_config.json does exist. So in this script we tweak existing kubelet_config.json then we let bootstrap to start kubelet with any kubelet_extra_args.

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: Agree. Per our conversation 2 days ago, we should only yum if it is not a custom ami.

cdk/domino_cdk/eks_stack.py
```diff
@@ -400,17 +402,78 @@ def provision_eks_iam_policies(self):
...
+                            ),
+                        ),
+                    ),
+                )
+            if isinstance(ud, ec2.UserData):
+                mime_user_data.add_part(ec2.MultipartBody.from_user_data(ud))
+            proceed = True
+
+        if not proceed:
+            return None
```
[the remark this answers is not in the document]
ddl-denis-parnovskiy: I think you are right. This is a leftover of previous logic in this function that did handle only user_data and did not care about ami.

cdk/domino_cdk/eks_stack.py
```diff
@@ -425,20 +488,23 @@ def provision_managed_nodegroup(self, name: str, cfg: dict, max_nodegroup_azs: i
...
+            else:
+                disk_size = cfg["disk_size"]
+                lts = None
 
             key_name = cfg.get("key_name", None)
             self.cluster.add_nodegroup_capacity(
-                f"{name}-{i}",  # this might be dangerous
-                nodegroup_name=f"{self.name}-{name}-{az}",  # this might be dangerous
+                f"{self.name}-{name}-{i}",  # this might be dangerous
+                nodegroup_name=f"{self.name}-{name}-{i}",
```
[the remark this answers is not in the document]
ddl-denis-parnovskiy: ok

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: ok

cdk/domino_cdk/eks_stack.py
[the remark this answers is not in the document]
ddl-denis-parnovskiy: EKS adds its logic as LAST part.

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: I'll try to change ASG{i} to indexed_name and see how it lives

cdk/domino_cdk/eks_stack.py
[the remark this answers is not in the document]
ddl-denis-parnovskiy: unfortunately, cannot use az for id because az is a token.

I have changed for the name, though

See https://us-west-2.console.aws.amazon.com/ec2autoscaling/home?region=us-west-2#/details/denis-test1-nvidia-us-west-2a?view=details

Also, please be aware, that for the managed groups EKS takes the provided group as a template and creates a new group with the same parameters but random ID. I believe this is part of managing, it clones what we have provided and changes it to the current policies.

See https://us-west-2.console.aws.amazon.com/ec2autoscaling/home?region=us-west-2#/details/eks-dabcff72-4cc4-d1c4-8d6d-73dfb6b5d904?view=details

cdk/domino_cdk/eks_stack.py
ddl-denis-parnovskiy: Works fine. ID is not actually visible anywhere .

See https://us-west-2.console.aws.amazon.com/ec2autoscaling/home?region=us-west-2#/details/denis-test1-compute-us-west-2a?view=details

Logical ID UnmanagedNodeGroupcomputedenistest1compute0ASG5CC272DB appears in tags so id field is not entorely lost

cdk/domino_cdk/eks_stack.py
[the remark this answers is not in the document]
Secretions: Got it

Secretions (approved):

PR #24, opened 2021-06-14
CLI features (schema upgrading, iam policy generator) and misc fixes

Secretions:
* Add `util.py` program to run misc management functions, remove existing ones from `app.py`
* Create a composable template rather than use the hardcoded one
* Add a config file loader/linter that can also dump the loaded config to text (ie for upgrading to current schema)
* Ensure all important outputs are on the scope of the main stack (ie bastionpublicip retrievable from primary stack rather than having to go into vpc sub-stack)
* Create IAM policy generator for both cdk synth and manual/terraform contexts
* Make everything terraform generates taggable
* Disable ingress/egress for default security group (compliance ask that I happened to be in the right place to do...)
* Remove implicit suffix from stack names (ie "domino" rather than "domino-eks-stack")
* Add option to provision a locally-stored IAM policy and use that for deployment to our terraform
* Add templted spot to add ACM cert ARN (tho perhaps this should be a proper config?)
* Integration test user moved to dev defaults values in template generator (also added replica setting)
* Made IMDS v2 (metadata v2) optional for now. 4.5.1 didn't support it (though this was just merged), and seems like having a safety valve is probably the right move to start out with
* Update IAM policies for tags can be updated

Also some cleanup followups from the object-based config:
* In provisioning contexts, import the domio_cdk.config module as config rather than individual classes (ie S3, EFS, etc.)
  * This disambiguates what the objects are, as previously `efs` was the `aws_cdk.aws_efs` module and `EFS` was the `EFS` config representation object from `domino_cdk.config.efs`
* Removes the `machine_image` block and folds its two attributes (`ami_id` and `user_data`) into the nodegroups
* Also remove the `_get_machine_image()` helper function, which does nearly nothing at this point


cdk/domino_cdk/config/eks.py
```diff
@@ -29,12 +29,15 @@ class NodegroupBase:
...
-                       setup your custom AMI or your user_data script _MANUALLY_ to join the cluster.
-                       Additionally, options that use user_data (ie ssm_agent) are no longer valid.
-                       machine_image:
-                         ami_id: ami-123abc
-                         user_data: ...
+        ami_id: ami-123abc - AMI to use for nodegroup, empty/null will use the default EKS AMI, which
+                             is the default. When specifying an AMI, you MUST specify a custom user_data
+                             script to join the node to the cluster, and this script must do any sort of
+                             node setup that is desired. Options that are implemented via user_data (ie
+                             labels, taints, ssm_agent) are not supported with custom AMIs, and must be
```
ddl-denis-parnovskiy, line 36: This reads as "user_data options are not supported with custom AMI"

-> "Labels, taints and ssm_agent config options are not supported with custom AMIs, and must be implemented manually via user_data"

cdk/domino_cdk/config/eks.py
```diff
@@ -29,12 +29,15 @@ class NodegroupBase:
...
+                             implemented manualyl.
+        user_data: ... - Custom script for user_data, ran by cloud_init on node startup. When using the
+                         default EKS AMI, this will be ran in addition to the default user_data scripts.
```
ddl-denis-parnovskiy, line 39: "in addition" -> "before"

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
```diff
@@ -159,7 +160,7 @@ def provision_unmanaged_nodegroup(self, name: str, ng: Type[EKS.NodegroupBase],
             ).items():
                 cdk.Tags.of(asg).add(str(k), str(v), apply_to_launched_instances=True)
 
-            mime_user_data = self._handle_user_data(name, ami_id, ng.ssm_agent, [asg.user_data, user_data])
+            mime_user_data = self._handle_user_data(name, ng.ami_id, ng.ssm_agent, [asg.user_data, ng.user_data])
```
ddl-denis-parnovskiy, line 163: my bad.

in managed NG, EKS adds generated code AFTER user_data.

here I add EKS-generated code BEFORE user_data.

We better be consistent. 

E.g. we change kubelet config in use_data and EKS code starts kubelet.

If we execute EKS code first, our changes will have no effect.

cdk/domino_cdk/config/iam.py
```diff
@@ -0,0 +1,400 @@
...
+            "Condition": {"ForAnyValue:StringEquals": {"aws:CalledVia": ["cloudformation.amazonaws.com"]}},
+            "Resource": [f"arn:aws:ecr:*:{aws_account_id}:repository/{stack_name}*"],
+        }
+    ]
+
+    if not manual:
+        ecr = []
+
+    if use_bastion:
+        bastion = [
```
ddl-denis-parnovskiy, line 249: manual case heavily uses "*Resource" wildcards in part because otherwise the resulting policy with all GetResource, PutResource, DeleteResource, DescribeResources, ListResources, etc will not fit into default 6Kb policy size limit.

Since we have some additions here, I recommend to test with some long stack name and see if it still fits.

Or UT.

cdk/domino_cdk/config/eks.py
Secretions: I was trying to describe the reason the options are not supported. I updated the language to try and convey it without any room for ambiguity.

cdk/domino_cdk/config/eks.py
Secretions: I didn't mean to suggest order, just indicate that your custom user data script doesn't replace the default one when you don't have a custom AMI. It's probably good to mention that it's before so people know what to expect, though, so I updated the language to convey both things.

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
Secretions: So I should switch the last arg to `[ng.user_data, asg.user_data]`?

cdk/domino_cdk/config/iam.py
Secretions: I'll try a longer name for the moment and hope it still works, as if it works I'd like to just get it in for now. Whether it ends up being now or later, maybe the right thing to do is just split it into two policies. That's what we did for the old rancher-focused version of it.

cdk/domino_cdk/config/iam.py
```diff
@@ -0,0 +1,399 @@
+# Future TODO item: Incorporate IAM reqs into the provisioning
```
ddl-denis-parnovskiy, line 1: Based on recent reviews and confusion, I would greatly recommend to keep original policies in JSON files and just do replace here. 

It is extremely hard to understand what is going on here and I do not understand the value of this move of the simple config files into complex code.

Secretions: The original policies still live in those files for now. This generator code exists for the moment as a tool to unify the policies more. We'll probably have to split it into two policies so it can all fit, which should enable us to get very close to having a single policy (maybe just the "coming from cloudformation" part being the only variation). After that point, this should be a lot less confusing.

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
ddl-denis-parnovskiy: yes

cdk/domino_cdk/config/iam.py
ddl-denis-parnovskiy: Your genius solution to split a policy into 2 policies and avoid 6K limit and use the same logic in both, is closing this.

cdk/Pipfile
```diff
@@ -4,24 +4,27 @@ verify_ssl = true
...
 "aws-cdk.aws-lambda" = "~=1.105.0"
-"aws-cdk.core" = "~=1.105.0"
 "aws-cdk.aws-kms" = "~=1.105.0"
-pyyaml = "*"
-"aws-cdk.aws-eks" = "~=1.105.0"
-requests = "*"
 "aws-cdk.aws-backup" = "~=1.105.0"
-"aws-cdk.aws-iam" = "~=1.105.0"
-"aws-cdk.aws-ecr" = "~=1.105.0"
+requests = "~=2.25.1"
```
ddl-denis-parnovskiy, line 17: Not sure why some packages are in "" and some are not. 

This list better be sorted.

cdk/domino_cdk/config/eks.py
```diff
@@ -30,12 +30,17 @@ class NodegroupBase:
...
         min_size: 1 - Minimum node count for nodegroup. Can't be 0 on managed nodegroups.
         max_size: 10 - Maximum limit for node count in node gorup
-        machine_image: AMI and/or user_data script for nodes in nodegroup. When overriding AMI, you MUST
-                       setup your custom AMI or your user_data script _MANUALLY_ to join the cluster.
-                       Additionally, options that use user_data (ie ssm_agent) are no longer valid.
-                       machine_image:
-                         ami_id: ami-123abc
-                         user_data: ...
+        ami_id: ami-123abc - AMI to use for nodegroup, empty/null will use the default EKS AMI, which
+                             is the default. When specifying an AMI, you MUST specify a custom user_data
```
ddl-denis-parnovskiy, line 34: "the default EKS AMI, which is the default" -> "the default EKS AMI"

cdk/domino_cdk/config/eks.py
```diff
@@ -30,12 +30,17 @@ class NodegroupBase:
...
+                             script to join the node to the cluster, and this script must do any sort of
+                             node setup that is desired. Additionally, some nodegroup options (ie labels,
+                             taints, ssm_agent) depend on the default user data and are not supported when
+                             using custom user_data. When providing custom user_data, you must implement
```
ddl-denis-parnovskiy, line 38: "depend on the default user data" -> user_data

Question: if the customer writes a script that 100% matches bootstrap.sh, can they use labels-etc?

cdk/domino_cdk/config/iam.py
```diff
@@ -0,0 +1,412 @@
...
+                "ec2:*Vpc*",
+                "s3:GetBucketLocation",
+                "s3:GetObject",
+                "s3:ListBucket",
+            ]
+        )
+        general["Resource"] = ["*"]
+    general["Action"] = sorted(general["Action"])
+
+    base_iam = {
```
ddl-denis-parnovskiy, line 390: Where is the genius approach to pack permissions into 2 policies?

ddl-denis-parnovskiy (commented):
Some nits

cdk/domino_cdk/config/iam.py
Secretions: That was going to be future work, and not part of this PR, some PR that merges the policies and then splits them. This was step 1. However, Steve already hit the "long name" test you'd mention, so I will probably make the two files/two policies before I merge this (but the merging of the policies will still be followup work).

cdk/domino_cdk/config/eks.py
Secretions: No, because we wouldn't know whether it was the exact same script. If we don't know what to expect, we're hands-off.

Secretions: "Depend on the default user data" *is* what I wanted to get across--they depend on the assumption that the user_data script is the default one.

ddl-denis-parnovskiy: > No, because we wouldn't know

The question is not about what we know. If they provide image that has all the features of original EKS image and provide user_data that is identical to the original EKS user data (this is pretty valid scenario, most of the people do not build AMI from scratch, they take the ready one and add what they need), then labels would work. 

We just do not allow it because we cannot be sure they have all the features in AMI.

But the explanation "why" makes if more confusing. So just say "don't" and this is it.

"Additionally, some nodegroup options (ie labels, taints, ssm_agent) depend on the default user data and are not supported when using custom user_data"
-> "The nodegroup options like labels, taints and ssm_agent are not supported when using custom user_data."

Secretions: I really want to get the why in there. I gave it another shot, trying to simplify the wording. Please take another look.

cdk/Pipfile
Secretions: I think it does it for packages that have the "." in them.

cdk/domino_cdk/config/eks.py
Secretions: I updated the phrasing.

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
Secretions: Changed this.

cdk/domino_cdk/config/iam.py
Secretions: I implemented the split part.

cdk/Pipfile
ddl-denis-parnovskiy: No. "" is not required for the package names. I have tried and all works without "". Please make consistent.

cdk/domino_cdk/config/eks.py
ddl-denis-parnovskiy: The new wording feels good to me. Closing this.

cdk/domino_cdk/config/eks.py
```diff
@@ -30,12 +30,17 @@ class NodegroupBase:
         key_name: some-key-pair - Pre-existing AWS key pair to configure for instances in the nodegorup
         min_size: 1 - Minimum node count for nodegroup. Can't be 0 on managed nodegroups.
         max_size: 10 - Maximum limit for node count in node gorup
-        machine_image: AMI and/or user_data script for nodes in nodegroup. When overriding AMI, you MUST
-                       setup your custom AMI or your user_data script _MANUALLY_ to join the cluster.
-                       Additionally, options that use user_data (ie ssm_agent) are no longer valid.
-                       machine_image:
-                         ami_id: ami-123abc
-                         user_data: ...
+        ami_id: ami-123abc - AMI to use for nodegroup, empty/null will default to the current EKS AMI,
```
ddl-denis-parnovskiy, line 33: > AMI to use for nodegroup, empty/null will default to the current EKS AMI,

Comma -> dot

cdk/Pipfile
ddl-denis-parnovskiy: Per slack conversation: just delete Pipfile and Pipfile.lock since we have all in setup.py and it is called anyway when doing bla bla requirements.txt

ddl-denis-parnovskiy: Please sort name list in setup.py

Secretions: Works for me.

ddl-denis-parnovskiy (approved):

PR #25, opened 2021-06-15
Modularize/Clean-up Provisioning

Secretions:
We've been implementing this thus far as both one single giant stack, as well as one single giant python file. This splits each concern into its own python module, and makes each module optionally deployable as a nested stack.

The reason for the optionality of the nested stacks is that there are limits of 200 for both the amount of resources in a given stack, as well as 200 stacks in a given account. A recent test install had about 140 resources, and while I don't see us increasing the resource count significantly from here on, that's roughly 75% of the way there.

Likewise, modularizing everything resulted in 7 separate stacks, giving us a limit of about 28 deployments.

The code for a nested module is 95% the same, so I just made this optional so people can optimize for whichever is the greater concern in their accounts. The default configuration will be modularizing EKS, while leaving the rest of the resources in the default stack, providing a good middle ground (eks has the most things in it).

Other than the minor changes listed below, the actual provisioning code is unchanged from the original beyond what was needed for modularization (ie changing `self` to `self.scope`, etc.) and splitting up one or two methods (into otherwise identical code).

Those additional minor changes:

* Automatically add autoscaler label hints
* Move availability_zones setting into VPC section, as it's really vpc-specific (which is a fact made more clear after the modularization)
  * I did _not_ make a new schema version
* Optimized out identical per-az launch templates...
  * Managed, straightforward: https://github.com/dominodatalab/cdk-cf-eks/commit/977eaada41883a35f93ee7ee2889e24e9134e20b
  * Unmanaged, has a questionable aspect: https://github.com/dominodatalab/cdk-cf-eks/commit/503e5b637712266521da1055ec6abd6138c3498a
    * The ASG is used in setting up the user data, which means the ASG for the first AZ is used to create the launch template for that unmanaged nodegroup. The actual content isn't AZ-specific, but it's gotten from an object created for the first AZ. This is a little unclean, but since it still should be deterministically correct I think it's less bad than making identical launch templates for no reason.


ddl-denis-parnovskiy (approved):
About unmanaged groups approach.

user data in the template is added by CDK itself. If there is nothing az-specific, we are good.

Of course this can change in next release on CDK but I do not foresee that kubelet init code there will be somehow specific to the AZ except the tags (which we can ignore)

cdk/domino_cdk/provisioners/eks.py
```diff
@@ -0,0 +1,518 @@
...
+
+manifests = [
+    [
+        "calico",
+        "https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.7.10/config/master/calico.yaml",
+    ]
+]
+
+
+class DominoEksProvisioner:
```
ddl-denis-parnovskiy, line 27: May be rename EKS to Cluster in this file?

Here we have a mix of EKS, IAM, EC2, IAM/S3 and IAM/ECR. Plus bastion and Calico.

Secretions: Tell me what you think of the latest commit. I pulled Calico out into its own thing, and made the cluster, iam provisioning and nodegroup provisioning their own submodules where it's more distinct what role they each play in provisioning the eks cluster and how they each linearly input into each other ("provision the control plane" -> "create the IAM role for the nodegroups" -> "provision the worker nodes").

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
```diff
@@ -1,265 +1,52 @@
...
-            iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKSWorkerNodePolicy'),
-            iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEC2ContainerRegistryReadOnly'),
-            iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKS_CNI_Policy'),
-            iam.ManagedPolicy.from_aws_managed_policy_name('AmazonSSMManagedInstanceCore'),
-        ]
-        if r53_zone_ids:
-            managed_policies.append(self.route53_policy)
+        def provision_nodegroup(nodegroup: EKS.NodegroupBase, prov_func):
+            for name, ng in nodegroup.items():
+                if not ng.machine_image or not ng.machine_image.ami_id:
```
ddl-denis-parnovskiy, line 39: I did it in my change but forgot to update the template.

ami_id and user_data become independent values.

we need to eliminate "machine_image" layer and more ami_id and user_data up.

cdk/domino_cdk/provisioners/eks.py
ddl-denis-parnovskiy: I like it

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
```diff
@@ -0,0 +1,278 @@
...
+
+                if taints := ng.taints:
+                    extra_args.append(
+                        "--register-with-taints={}".format(",".join(["{}={}".format(k, v) for k, v in taints.items()]))
+                    )
+                options["bootstrap_options"] = eks.BootstrapOptions(kubelet_extra_args=" ".join(extra_args))
+
+            self.cluster.connect_auto_scaling_group_capacity(asg, **options)
+
+    def _get_machine_image(self, cfg_name: str, image: MachineImage) -> Tuple[Optional[str], Optional[str]]:
```
ddl-denis-parnovskiy, line 230: I do not welcome this change to passing MachineImage instead of config.

Default EKS AMI doesn't have meaningful user_data. EKS adds user_data directly to the launch template (via nodegroup). Adding arbitrary user_data to machine image object and then extracting this user_data from machine image just to add it to a LaunchTemplate is counterproductive and subject to side effects.

I recommend to keep ami_id and user_data separate until they are combined into a launch template.

Currently this code breaks user_data handling because machine_image is created without user_data (in provision_xxx_nodegroup second line) and here we are getting zilch.

Secretions: There might be a misunderstanding here, the `MachineImage` object being passed in here is the object from the config, _not_ a CDK machine image object. Maybe I should do some followup work to disambiguate/clearly namespace our config objects from aws api calls/results.

Secretions: Weird, I commented on this, it showed up twice, I deleted the second iteration of the comment, reloaded, and now it's gone entirely. Lovely. Guessing the two comments was just a display error. Anyway...

I think there might be a misunderstanding, the `image` variable being passed here's type, `MachineImage`, is the domino_cdk config object (ie just a container for the config), and *NOT* an aws cdk representation of a machine image. So we are still passing the config, but we're just passing the relevant portion of the config.

I think maybe I need to do some followup work to the object-based config to disambiguate/namespace the config objects so they don't get confused with aws_cdk things (like, as imported, `EKS` is a config object but `eks` is the aws_cdk's eks api... which is confusing). I will get that into my next PR.

ddl-denis-parnovskiy: Agree. Confused with IMachineImage

ddl-denis-parnovskiy: See my previous comment.

How about eliminating MachineImage and _get_machine_image (it doesn't do much anymore).

user_data is launch template concept, not an image concept.

Secretions: I was already planning to get rid of `_get_machine_image` in followup, so I will handle all three things (getting rid of the machine image config block, removing _get_machine_image, and disambiguating the config variables) in the followup.

ddl-denis-parnovskiy:
No more comments

PR #26, opened 2021-06-19
Adding IAM roles for K8s service accounts (PLAT-2264)

ddl-denis-parnovskiy:
We configure IAM roles to associate with service accounts in k8s. 

We have a switch to choose between the node level roles and service level nodes.

ECR permissions for the nodes are changed to the recommended for the worker nodes.

See code comments for more details.

### DIFF
--- a/cdk/domino_cdk/aws_configurator.py
+++ b/cdk/domino_cdk/aws_configurator.py
@@ -22,13 +22,10 @@ manifests = [
 # deprovisoning efs backups/route53, tagging the eks cluster until
 # the CloudFormation api supports it, etc.)
 class DominoAwsConfigurator:
-    def __init__(
-        self, scope: cdk.Construct, eks_cluster: eks.Cluster, vpc: ec2.Vpc, s3_api_statement: iam.PolicyStatement
-    ):
+    def __init__(self, scope: cdk.Construct, eks_cluster: eks.Cluster, vpc: ec2.Vpc):
         self.scope = scope
         self.eks_cluster = eks_cluster
         self.vpc = vpc
-        self.s3_api_statement = s3_api_statement
 
         self.install_calico()
 
--- a/cdk/domino_cdk/aws_configurator.py
+++ b/cdk/domino_cdk/aws_configurator.py
@@ -4,7 +4,6 @@ from re import split as re_split
 
 import aws_cdk.aws_ec2 as ec2
 import aws_cdk.aws_eks as eks
-import aws_cdk.aws_iam as iam
 from aws_cdk import core as cdk
 from requests import get as requests_get
 from yaml import safe_load as yaml_safe_load
--- a/cdk/domino_cdk/provisioners/eks/eks_iam.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_iam.py
@@ -123,7 +123,7 @@ class DominoEksIamProvisioner:
                         "s3:ListMultipartUploadParts",
                         "s3:AbortMultipartUpload",
                     ],
-                    resources={f"{b.bucket_arn}*" for b in buckets.values()},
+                    resources=[f"{b.bucket_arn}*" for b in buckets.values()],
                 ),
             ],
         )
--- a/cdk/domino_cdk/provisioners/eks/eks_iam.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_iam.py
@@ -69,26 +69,7 @@ class DominoEksIamProvisioner:
         ]
 
         if r53_zone_ids:
-            managed_policies.append(
-                iam.ManagedPolicy(
-                    self.scope,
-                    "route53",
-                    managed_policy_name=f"{stack_name}-route53",
-                    statements=[
-                        iam.PolicyStatement(
-                            actions=["route53:ListHostedZones"],
-                            resources=["*"],
-                        ),
-                        iam.PolicyStatement(
-                            actions=[
-                                "route53:ChangeResourceRecordSets",
-                                "route53:ListResourceRecordSets",
-                            ],
-                            resources=[f"arn:aws:route53:::hostedzone/{zone_id}" for zone_id in r53_zone_ids],
-                        ),
-                    ],
-                )
-            )
+            managed_policies.append(self.provision_r53_policy(stack_name, r53_zone_ids))
 
         if buckets:
             managed_policies.append(self.provision_node_s3_iam_policy(stack_name, buckets))
@@ -101,11 +82,31 @@ class DominoEksIamProvisioner:
             managed_policies=managed_policies,
         )
 
-    def provision_node_s3_iam_policy(self, name: str, buckets: Dict[str, Bucket]) -> iam.ManagedPolicy:
+    def provision_r53_policy(self, stack_name: str, r53_zone_ids: List[str]):
+        return iam.ManagedPolicy(
+            self.scope,
+            "route53",
+            managed_policy_name=f"{stack_name}-route53",
+            statements=[
+                iam.PolicyStatement(
+                    actions=["route53:ListHostedZones"],
+                    resources=["*"],
+                ),
+                iam.PolicyStatement(
+                    actions=[
+                        "route53:ChangeResourceRecordSets",
+                        "route53:ListResourceRecordSets",
+                    ],
+                    resources=[f"arn:aws:route53:::hostedzone/{zone_id}" for zone_id in r53_zone_ids],
+                ),
+            ],
+        )
+
+    def provision_node_s3_iam_policy(self, stack_name: str, buckets: Dict[str, Bucket]) -> iam.ManagedPolicy:
         return iam.ManagedPolicy(
             self.scope,
             "S3",
-            managed_policy_name=f"{name}-S3",
+            managed_policy_name=f"{stack_name}-S3",
             statements=[
                 iam.PolicyStatement(
                     actions=[
--- a/cdk/domino_cdk/provisioners/eks/eks_iam.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_iam.py
@@ -11,7 +11,7 @@ class DominoEksIamProvisioner:
     ) -> None:
         self.scope = scope
 
-    def provision(self, name: str, cluster_name: str, s3_policy: iam.ManagedPolicy, r53_zone_ids: List[str]):
+    def provision(self, name: str, cluster_name: str, r53_zone_ids: List[str]):
         asg_group_statement = iam.PolicyStatement(
             actions=[
                 "autoscaling:DescribeAutoScalingInstances",
@@ -62,12 +62,16 @@ class DominoEksIamProvisioner:
 
         ecr_policy = iam.ManagedPolicy(
             self.scope,
-            "DominoEcrReadOnly",
+            f"{name}-DominoEcrRestricted",
             managed_policy_name=f"{name}-DominoEcrRestricted",
             statements=[
                 iam.PolicyStatement(
                     effect=iam.Effect.DENY,
-                    actions=["ecr:*"],
+                    actions=[
+                        "ecr:BatchCheckLayerAvailability",
+                        "ecr:BatchGetImage",
+                        "ecr:GetDownloadUrlForLayer",
+                    ],
                     conditions={"StringNotEqualsIfExists": {"ecr:ResourceTag/domino-deploy-id": name}},
                     resources=[f"arn:aws:ecr:*:{self.scope.account}:*"],
                 ),
@@ -75,7 +79,6 @@ class DominoEksIamProvisioner:
         )
 
         managed_policies = [
-            s3_policy,
             ecr_policy,
             autoscaler_policy,
             iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKSWorkerNodePolicy'),
--- a/cdk/domino_cdk/provisioners/eks/eks_iam.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_iam.py
@@ -1,7 +1,8 @@
-from typing import List
+from typing import List, Dict
 
 import aws_cdk.aws_iam as iam
 from aws_cdk import core as cdk
+from aws_cdk.aws_s3 import Bucket
 
 
 class DominoEksIamProvisioner:
@@ -11,7 +12,7 @@ class DominoEksIamProvisioner:
     ) -> None:
         self.scope = scope
 
-    def provision(self, name: str, cluster_name: str, r53_zone_ids: List[str]):
+    def provision(self, stack_name: str, cluster_name: str, r53_zone_ids: List[str], buckets: Dict[str, Bucket]):
         asg_group_statement = iam.PolicyStatement(
             actions=[
                 "autoscaling:DescribeAutoScalingInstances",
@@ -25,7 +26,7 @@ class DominoEksIamProvisioner:
         autoscaler_policy = iam.ManagedPolicy(
             self.scope,
             "autoscaler",
-            managed_policy_name=f"{name}-autoscaler",
+            managed_policy_name=f"{stack_name}-autoscaler",
             statements=[
                 iam.PolicyStatement(
                     actions=[
@@ -40,30 +41,10 @@ class DominoEksIamProvisioner:
             ],
         )
 
-        if r53_zone_ids:
-            route53_policy = iam.ManagedPolicy(
-                self.scope,
-                "route53",
-                managed_policy_name=f"{name}-route53",
-                statements=[
-                    iam.PolicyStatement(
-                        actions=["route53:ListHostedZones"],
-                        resources=["*"],
-                    ),
-                    iam.PolicyStatement(
-                        actions=[
-                            "route53:ChangeResourceRecordSets",
-                            "route53:ListResourceRecordSets",
-                        ],
-                        resources=[f"arn:aws:route53:::hostedzone/{zone_id}" for zone_id in r53_zone_ids],
-                    ),
-                ],
-            )
-
         ecr_policy = iam.ManagedPolicy(
             self.scope,
-            f"{name}-DominoEcrRestricted",
-            managed_policy_name=f"{name}-DominoEcrRestricted",
+            f"{stack_name}-DominoEcrRestricted",
+            managed_policy_name=f"{stack_name}-DominoEcrRestricted",
             statements=[
                 iam.PolicyStatement(
                     effect=iam.Effect.DENY,
@@ -72,7 +53,7 @@ class DominoEksIamProvisioner:
                         "ecr:BatchGetImage",
                         "ecr:GetDownloadUrlForLayer",
                     ],
-                    conditions={"StringNotEqualsIfExists": {"ecr:ResourceTag/domino-deploy-id": name}},
+                    conditions={"StringNotEqualsIfExists": {"ecr:ResourceTag/domino-deploy-id": stack_name}},
                     resources=[f"arn:aws:ecr:*:{self.scope.account}:*"],
                 ),
             ],
@@ -86,13 +67,66 @@ class DominoEksIamProvisioner:
             iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKS_CNI_Policy'),
             iam.ManagedPolicy.from_aws_managed_policy_name('AmazonSSMManagedInstanceCore'),
         ]
+
         if r53_zone_ids:
-            managed_policies.append(route53_policy)
+            managed_policies.append(
+                iam.ManagedPolicy(
+                    self.scope,
+                    "route53",
+                    managed_policy_name=f"{stack_name}-route53",
+                    statements=[
+                        iam.PolicyStatement(
+                            actions=["route53:ListHostedZones"],
+                            resources=["*"],
+                        ),
+                        iam.PolicyStatement(
+                            actions=[
+                                "route53:ChangeResourceRecordSets",
+                                "route53:ListResourceRecordSets",
+                            ],
+                            resources=[f"arn:aws:route53:::hostedzone/{zone_id}" for zone_id in r53_zone_ids],
+                        ),
+                    ],
+                )
+            )
+
+        if buckets:
+            managed_policies.append(
+                self.provision_node_s3_iam_policy(stack_name, buckets)
+            )
+
 
         return iam.Role(
             self.scope,
-            f'{name}-NG',
-            role_name=f"{name}-NG",
+            f'{stack_name}-NG',
+            role_name=f"{stack_name}-NG",
             assumed_by=iam.ServicePrincipal('ec2.amazonaws.com'),
             managed_policies=managed_policies,
+        )
+
+    def provision_node_s3_iam_policy(self, name: str, buckets: Dict[str, Bucket]) -> iam.ManagedPolicy:
+        return iam.ManagedPolicy(
+            self.scope,
+            "S3",
+            managed_policy_name=f"{name}-S3",
+            statements=[
+                iam.PolicyStatement(
+                    actions=[
+                        "s3:ListBucket",
+                        "s3:GetBucketLocation",
+                        "s3:ListBucketMultipartUploads",
+                    ],
+                    resources=["*"],
+                ),
+                iam.PolicyStatement(
+                    actions=[
+                        "s3:PutObject",
+                        "s3:GetObject",
+                        "s3:DeleteObject",
+                        "s3:ListMultipartUploadParts",
+                        "s3:AbortMultipartUpload",
+                    ],
+                    resources={f"{b.bucket_arn}*" for b in buckets.values()},
+                ),
+            ],
         )
--- a/cdk/domino_cdk/provisioners/eks/eks_iam.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_iam.py
@@ -1,4 +1,4 @@
-from typing import List, Dict
+from typing import Dict, List
 
 import aws_cdk.aws_iam as iam
 from aws_cdk import core as cdk
@@ -91,10 +91,7 @@ class DominoEksIamProvisioner:
             )
 
         if buckets:
-            managed_policies.append(
-                self.provision_node_s3_iam_policy(stack_name, buckets)
-            )
-
+            managed_policies.append(self.provision_node_s3_iam_policy(stack_name, buckets))
 
         return iam.Role(
             self.scope,
--- a/cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
@@ -61,16 +61,16 @@ ecr_write_permisions = [
 # Roles. The roles are the collection of the policies.
 roles = {
     # E.g. nucleus needs this role
-    "write_blobs_read_logs": [
+    "blobs-write--logs-read": [
         "blobs:write",
         "logs:read",
     ],
     # E.g. executor needs this role
-    "write_blobs": [
+    "blobs-write": [
         "blobs:write",
     ],
     # E.g. builder needs this role
-    "write_images": [
+    "images-write": [
         "ecr:write",
     ],
 }
--- a/cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
@@ -1,0 +1,176 @@
+from typing import Dict, List
+
+import aws_cdk.aws_eks as eks
+import aws_cdk.aws_iam as iam
+from aws_cdk import core as cdk
+from aws_cdk.aws_s3 import Bucket
+
+# Permission groups
+
+s3_write_permissions = [
+    "s3:PutObject",
+    "s3:DeleteObject",
+    "s3:ListMultipartUploadParts",
+    "s3:AbortMultipartUpload",
+]
+
+s3_read_permissions = [
+    "s3:GetObject",
+]
+
+ecr_policies = {
+    "ecr_read_write_permissions": [
+        "ecr:PutImageTagMutability",
+        "ecr:StartImageScan",
+        "ecr:ListTagsForResource",
+        "ecr:UploadLayerPart",
+        "ecr:BatchDeleteImage",
+        "ecr:ListImages",
+        "ecr:DeleteRepository",
+        "ecr:CompleteLayerUpload",
+        "ecr:TagResource",
+        "ecr:DescribeRepositories",
+        "ecr:DeleteRepositoryPolicy",
+        "ecr:BatchCheckLayerAvailability",
+        "ecr:ReplicateImage",
+        "ecr:GetLifecyclePolicy",
+        "ecr:PutLifecyclePolicy",
+        "ecr:DescribeImageScanFindings",
+        "ecr:GetLifecyclePolicyPreview",
+        "ecr:CreateRepository",
+        "ecr:PutImageScanningConfiguration",
+        "ecr:GetDownloadUrlForLayer",
+        "ecr:DeleteLifecyclePolicy",
+        "ecr:PutImage",
+        "ecr:UntagResource",
+        "ecr:BatchGetImage",
+        "ecr:DescribeImages",
+        "ecr:StartLifecyclePolicyPreview",
+        "ecr:InitiateLayerUpload",
+        "ecr:GetRepositoryPolicy",
+    ]
+}
+
+s3_policies = {
+    "write_blobs_read_logs": {
+        "blobs": s3_read_permissions + s3_write_permissions,
+        "logs": s3_read_permissions,
+    },
+    "write_blobs": {
+        "blobs": s3_read_permissions + s3_write_permissions,
+    },
+    "read_logs": {
+        "logs": s3_read_permissions,
+    },
+    "read_blobs": {
+        "blobs": s3_read_permissions,
+    },
+}
+
+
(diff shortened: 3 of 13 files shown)


Secretions (commented):
If we make any of version of this change, Domino won't work until we have things squared away on the agent side. So before everything else, I think we need to make this optional for the time being. Like if we tell it to provision iam service account stuff, it'll skip provisioning s3 iam for the node, otherwise it does the old behavior.

Didn't we also need to do some oidc thing for this to work?

I'd like to hardcode the IAM policies involved here and keep it out of the config file. I don't think we need that level of granularity of configuration, and it'll provide a lot of noise.

As for the overall approach, I'd like to _not_ provision the service accounts in CDK. Let's provision the policies for the service accounts in the DominoEksIamProvisioner and make it clear the deliverable here is just IAM roles.

So we'd have, in the cdk outputs, some variables like "blobs-bucket-role-arn".

Then in the fleetcommand_agent, we'd add an optional input for the iam roles, something like:

```
service_account_iam_roles:
  blobs_access: <blobs-bucket-role-arn from cdk output goes here>
  [...]
```

(and don't forget, we generate an agent template as an output as well, so we'd make sure it's in that too)

Then, when building the nucleus chart config in `installer/charts.py`, we'd feed this value to the nucleus chart, something like:

```
config:
  nucleus_iam_role: <role arn goes here>
```

Then, finally, when the nucleus chart creates its service account, it'll add the annotation for the iam role:

```
apiVersion: v1
kind: ServiceAccount
name: nucleus-service-account
metadata:
  annotations:
    eks.amazonaws.com/role-arn: <blobs bucket arn role>
```

And likewise, let's say the replicator needs the same iam role, you'd feed it the blobs bucket arn role as well and it'd add the annotation when creating its own service account, etc.

This way, the charts can continue to create their own service accounts, and the cdk doesn't need to know quite so much about the deployment (including namespaces), just what sets of IAM roles it needs to create.

ddl-denis-parnovskiy:
> make this optional for the time being

ok

> Didn't we also need to do some oidc thing for this to work?

Hmm. I usually test at least the happy path, if not the corner cases.

Here, CF/EKS provisions all required stuff automatically when you add a service account via CDK. It will even create an empty role if you did not provide one.

> I'd like to hardcode the IAM policies involved here and keep it out of the config file. 
> it'll provide a lot of noise.

I have tried this and I'm convinced that having list of permissions in the config makes it more clear and more flexible.

The code for the config is minimal. And I have an opposite opinion on the noise:

I can move the configuration part from the config file to the code as a dict in python code, but this widely considered to be an anti-pattern. And the code reduction will be literally removal of this part
```
    @dataclass
    class ServiceAccount:
        buckets: Dict[str, List[str]]
        ecr: List[str]

        @classmethod
        def load(cls, sa):
            return EKS.ServiceAccount(
                buckets = {
                              bucket_name: p
                              for bucket_name, p in sa.pop("s3_bucket_permissions").items()
                          },
                ecr = sa.pop("ecr_permissions"),
            )
```
plus a call to it.

This leads us to a discussion about customer-facing configs and internal configs. And this is very interesting discussion because it widely accepted in the industry that config change is a zillion times better than a code change and all data must be in the config/properties files and not in the constants in the code (and not in the literals FFS).

BTW, I have a similar sentiment about the nodegroups. Why we allow them to configure the names and numbers of the nodegroups? It is something that should not be configurable because it will break the agent workflow. They remove "platform" and create "management_cluster" and what we are going to do?

We should just hardcode nodegroups "number, labels, names" and "managed/unmanaged" choice and only have parameters
* size_of_compute_group: 4 medium 100Gb
* size_of_management_group: 1 medium 100Gb
* size_of_nvidia_group: 2 small 100Gb


> I'd like to not provision the service accounts in CDK.

Side note: In this case we need to provision OIDC and add it as trusted source in CDK. This is very non-trivial task.

In general: we do not provision "service accounts" in CDK. We provision IAM roles with fixed permissions and we create **handles** for these roles to be visible in k8s. Those handles are coming in a form of service accounts.

I see only 1 benefit of creating those handles later: 
* CDK agnostic of the namespaces.

But CDK is not agnostic of the namespaces now, because we do plan to create nodegroups for "platform" and "compute" in CDK.

A drawback of creating only roles in CDK: we either
* make new roles closely tailored to the current usage, making CDK non-agnosting of the services anyway.
* OR create a bunch or unused IAM roles hoping that we predict the future and if we need to change the permissions, we can pick one of the unused roles in the agent without need to re-provision.

Example: we have 3 S3 buckets. Not all the services need access to all 3 buckets.

What we do?
1. Create roles per current services?
2. Create 7 roles (for all subsets size up to 3 from set of 3)?

So a big conceptual questions are not about the accounts. 
It is about: 
* do we plan CDK to use names "platform" and "compute" or not. If we do, then arguments that CDK should not be aware about "platform" and "compute" for the accounts part are invalid.
* how exactly we make CDK agnostic of domino services and their permissions.

Secretions:
> > Didn't we also need to do some oidc thing for this to work?
> 
> Hmm. I usually test at least the happy path, if not the corner cases.
> 
> Here, CF/EKS provisions all required stuff automatically when you add a service account via CDK. It will even create an empty role if you did not provide one.

> (snipped from later) Side note: In this case we need to provision OIDC and add it as trusted source in CDK. This is very non-trivial task.

WRT what's required if we don't create service accounts, I suppose we could just create the regular default service account with an empty iam to force it to do the hard stuff (the oidc) for us?
 
> > I'd like to hardcode the IAM policies involved here and keep it out of the config file.
> > it'll provide a lot of noise.
> 
> I have tried this and I'm convinced that having list of permissions in the config makes it more clear and more flexible.
> 
> The code for the config is minimal. And I have an opposite opinion on the noise:

The code for the config is not minimal, there's more we'll have to add, and it's not stuff the people deploying are going to know anything about. They just need the iam roles to work, but won't know what is/isn't needed. Likewise, the iam roles rarely change.

We're ultimately going to need configs for blobs (once for nucleus, once for executor and once for replicator), the registry bucket for now, blobs+ecr for the builder, logs bucket for fluentd, route53 for external dns, some ec2 and/or autoscaling stuff for the cluster-autoscaler, backups bucket for the backup cron jobs, off the top of my head. This is a lot of noise in the config file for something the end user will never touch and isn't largely concerned with.
 
> I can move the configuration part from the config file to the code as a dict in python code, but this widely considered to be an anti-pattern. And the code reduction will be literally removal of this part
> 
> [...]
> 
> This leads us to a discussion about customer-facing configs and internal configs. And this is very interesting discussion because it widely accepted in the industry that config change is a zillion times better than a code change and all data must be in the config/properties files and not in the constants in the code (and not in the literals FFS).

If it's in some behind-the-scenes yaml properties file or something that's fine too, but I really just don't think any of this should be in the user-facing config file.

> BTW, I have a similar sentiment about the nodegroups. Why we allow them to configure the names and numbers of the nodegroups? It is something that should not be configurable because it will break the agent workflow. They remove "platform" and create "management_cluster" and what we are going to do?
> 
> We should just hardcode nodegroups "number, labels, names" and "managed/unmanaged" choice and only have parameters
> 
> * size_of_compute_group: 4 medium 100Gb
> * size_of_management_group: 1 medium 100Gb
> * size_of_nvidia_group: 2 small 100Gb

It (changing the name) should still work, the labels are what matters, and customers do have different customizations and preferences for the nodegroups. As well it can be useful for upgrades/transitions (ie create a second platform nodegroup, cordon nodes, move stuff over), something which I think is required for certain changes on the managed nodegroups. There is for sure some room for reducing or simplifying the config settings somewhat (for example, node-pool could be a setting rather than requiring the user to know the label beforehand), but I don't think it's right that we can simplify it to this degree. Regardless, I don't think it's an apples-to-apples comparison as, even if there is more configurability than needed, these nodegroups are something the user does still have visibility into and an opinion on the configuration of. The IAM policies are a behind-the-scenes thing that just needs to work.
 
> > I'd like to not provision the service accounts in CDK.
> 

> 
> In general: we do not provision "service accounts" in CDK. We provision IAM roles with fixed permissions and we create **handles** for these roles to be visible in k8s. Those handles are coming in a form of service accounts.
> 
> I see only 1 benefit of creating those handles later:
> 
> * CDK agnostic of the namespaces.
> 
> But CDK is not agnostic of the namespaces now, because we do plan to create nodegroups for "platform" and "compute" in CDK.
> 
> A drawback of creating only roles in CDK: we either
> 
> * make new roles closely tailored to the current usage, making CDK non-agnosting of the services anyway.
> * OR create a bunch or unused IAM roles hoping that we predict the future and if we need to change the permissions, we can pick one of the unused roles in the agent without need to re-provision.
> 
> Example: we have 3 S3 buckets. Not all the services need access to all 3 buckets.
> 
> What we do?
> 
> 1. Create roles per current services?
> 2. Create 7 roles (for all subsets size up to 3 from set of 3)?
> 
> So a big conceptual questions are not about the accounts.
> It is about:
> 
> * do we plan CDK to use names "platform" and "compute" or not. If we do, then arguments that CDK should not be aware about "platform" and "compute" for the accounts part are invalid.
> * how exactly we make CDK agnostic of domino services and their permissions.

CDK does not need to know much about the actual configuration of the deployment right now. Having to create some ASGs with particular labels is much less than knowing intended namespaces, installed services and their mappings.

Right now, CDK needs to know about your individual deployment:

* What autoscaling groups you want with what labels/taints

Additionally: The agent also has a reverse dependency that CDK has provisioned a single IAM role for the instances that covers the needs of all services.

It doesn't care about namespaces (which are not the same as labels--a couple compute things even live on platform nodes), it doesn't care about charts/services or their necessary service accounts or what namespaces they'll go in. IAM roles are statically defined/global.


If we just create statically defined roles, CDK still just needs to know:

* What autoscaling groups you want with what labels/taints

Additionally: The agent has a similar reverse dependency that CDK has provisioned a some IAM role for each type of access required for all services. It needs to know each individual role's arn to send to the charts.

It provides certain pre-defined IAM roles, and a given CDK deployment does need to provide all the IAM roles an install will use, but it doesn't need to know anything about how those will be deployed into kubernetes. It also doesn't care about charts/services or the service accounts they need or what namespaces in k8s they will go into.


If we create service accounts, CDK then needs to know:

* What autoscaling groups you want with what labels/taints
* The precise name the deployment will use (because this factors into creating the namespaces)
  * Yes, this usually will match the name in CDK, but it doesn't have to. If CDK is being provisioned without the deployer, this is more likely to be an issue.
  * It also means if the schema for naming the namespaces changes in anyway, CDK now needs to be updated correspondingly
* What namespaces you're going to use (which may or may not have anything to do with the label)
* What charts/services the fleetcommand agent will install
* The mapping of charts onto the namespaces to determine what service accounts need to be created
* The mapping of the roles into those service accounts

Additionally: The agent has has a reverse dependency that CDK has provisioned not only an IAM role for every type of access required, but an individual service account in a specific namespace for every service that can possibly consume that IAM role. It needs to know each service account that's been created to send to the charts.


The CDK installer is now tightly coupled with the Domino app installation. If I add new "blob-using service X" to the fleetcommand agent installer, I now also have to go into the CDK configs and add a whole new service for it, requiring the end user to reprovision with CDK in AWS. As where if we just deliver the roles, I can just setup the chart to use the existing blob access IAM role when it creates its service account. When we upgrade to this version, all that needs to happen is the fleetdcommand agent does an upgrade as per normal.

ddl-denis-parnovskiy:
Let's split this into independent parts.

### Move roles from customer-facing config to internal config

Agree. I do not care much if it is a config file with `{ "a": true }` or Python file with `c = { "a": True }` so let it be a Python file.


### Today CDK-agent contract doesn't care about namespaces "deployment_id-platform" and "deployment_id-compute", it only cares about labels "platform" and "compute".

> these nodegroups are something the user does still have visibility into and an opinion on the configuration of

OK. We still need to restrict and validate that:
* they do not use more than number of spaces we support. We do not want them to create 2 different nodegroups for "platform"
* they use both these labels. We do not want them to create a nodegroup for platform and this is it.

I do not remember if we have such validators and if this is documented.


### We can create some small set of universal roles that cover all use cases

I think you are missing my point, because you keep avoiding my examples of mix and match of the permissions and use your own example of "blob accessing service".

I'll ask again.
* We have 3 buckets. Blobs, logs and backups
* We have 1 service reading/writing blobs and reading logs
* We have 1 service reading/writing blobs 
* We have 1 service writing logs
* We have 1 service writing backups

What our "future-proof" set of services, that "256K enough for everyone", would be? There is quite a number of combinations, more than 7 if you count read and write as separate permissions.

What if tomorrow we add a service that will only read blobs and logs? 

The only feasible answer we have today is: we create roles as required today. And we go back and update CDK if a new role with a new combination of permissions is required.

Yes it means that if we do not create service accounts, we still have to go to CDK and add code for a new service. But not for every service, see below.

### If we create service accounts, CDK needs to know about the services and charts.

I gotcha. I was proposing that service accounts will be "service_account_for_role_A". I forgot that we have to have account per service in k8s.

This part is agreed upon.


## Resume
* IAM group config moves inside the app.
* We do not create accounts in CDK. It is my misconception, I was planning to create account per role, not per service.
* We need checks that the customers do not misuse labels platform and compute.
* We can not create a set of IAM roles that we will not need to change in the future, but this is irrelevant because there is no way around it; none of the proposals solves this problem. If we do not create service account for each service in CDK, then we do not need to change CDK for every service, only for the new roles.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,176 @@
...
+    def __init__(
+        self,
+        scope: cdk.Construct,
+    ) -> None:
+        self.scope = scope
+
+    def print_children(self, construct):
+        print(construct.node.path, construct.to_string())
+        for c in construct.node.children:
+            self.print_children(c)
```
Secretions, line 95: Is this just a debug thing?

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,176 @@
...
+            ),
+        )
+        statement_json['Condition']['StringLike'] = cdk.CfnJson(
+            self.scope, "OidcJson", value={f"{fn}:aud": "sts.amazonaws.com", f"{fn}:sub": "system:serviceaccount:*"}
+        )
+        managed_policies = {}
+        for name, cfg in ecr_policies.items():
+            managed_policies[name] = self.create_ecr_policy(stack_name, name, cfg)
+        for name, cfg in s3_policies.items():
+            managed_policies[name] = self.create_s3_policy(name, name, cfg, buckets)
```
Secretions, line 123: Is the first `name` supposed to be `stack_name`?

cdk/domino_cdk/aws_configurator.py
```diff
@@ -25,13 +25,10 @@
...
 # the CloudFormation api supports it, etc.)
 class DominoAwsConfigurator:
-    def __init__(
-        self, scope: cdk.Construct, eks_cluster: eks.Cluster, vpc: ec2.Vpc, s3_api_statement: iam.PolicyStatement
-    ):
+    def __init__(self, scope: cdk.Construct, eks_cluster: eks.Cluster, vpc: ec2.Vpc):
         self.scope = scope
         self.eks_cluster = eks_cluster
         self.vpc = vpc
-        self.s3_api_statement = s3_api_statement
```
Secretions, line 34: Mentioned this previously, but we do still need to make separate roles / node roles some sort of toggle for the time being.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,176 @@
...
+
+        for name, policy_list in roles.items():
+            iam_role = iam.Role(
```
Secretions, line 126: We are ultimately going to need to create CfnOutputs for the roles. It will be important to do it as part of the parent stack to make it sufficiently accessible, which is something I've added for my own branch, `cli-iam-tagging-2`. Might be worth rebasing on that so it's informed by those changes (or just wait for that to merge in before rebasing so you don't have to do it twice).

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,176 @@
...
+        "logs": s3_read_permissions,
+    },
+    "read_blobs": {
+        "blobs": s3_read_permissions,
+    },
+}
+
+
+# Roles
+roles = {
```
steved, line 72: I'm worried we're missing: fluentd (for S3 writing), cluster-autoscaler (for node management), docker-registry (for S3 storage), nucleus (for volume snapshotting), backups (git/mongo/postgres to S3), fc-agent (optional, but a few observability things). Since this is incremental work, I think it's ok, we just need to make sure we continue to incrementally move things to pod roles.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,176 @@
...
+    "nucleus": [
+        "write_blobs_read_logs",
```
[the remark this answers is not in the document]
ddl-denis-parnovskiy: > We should be able to make the permissions distinct policies (ie a blob s3 read policy, a blob s3 write policy, an s3 log read policy), assemble them into roles with with the functions in your s3 policies section (ie write_blobs_read_logs) 

This is exactly what I'm doing here.
1. Have blocks of the permissions.
2. Assemble them into the named roles.

>and then remove this service-aware level.

There is no service-aware level. Duno what you are talking about. These are the roles and role names. 

I name them after the usage examples I'm aware of. I.e. we need roles that are tailored for our usage. nucleus has distinctive usage, executor needs an obvious role we need to create, fluentd needs a distinctive role, builder needs a distinctive role. Perhaps there are some other examples, please add.

We can name the roles differently but it doesn't change the fact that we do have AT LEAST 4 services that dictate AT LEAST 4 distinctive roles to be created.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
ddl-denis-parnovskiy: Blindness.I though I have removed this debug thing.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,176 @@
...
+        managed_policies = {}
+        for name, cfg in ecr_policies.items():
+            managed_policies[name] = self.create_ecr_policy(stack_name, name, cfg)
+        for name, cfg in s3_policies.items():
+            managed_policies[name] = self.create_s3_policy(name, name, cfg, buckets)
+
+        for name, policy_list in roles.items():
+            iam_role = iam.Role(
+                self.scope,
+                f"{stack_name}-IAM-role-for-k8s-{name}",
```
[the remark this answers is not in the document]
ddl-denis-parnovskiy: There are not service-avare component.

But I agree, we can make the structure one level lower by not combining s3 and ecr permissions into one policy and just have separate s3 and ecr policies and combine them into roles.

I'll do this.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
ddl-denis-parnovskiy: No, it is policy name.

But create_s3_policy had unused second parameter. Thanks.

cdk/domino_cdk/aws_configurator.py
ddl-denis-parnovskiy: Working on this.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
ddl-denis-parnovskiy: OK, let's wait for your merge.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
ddl-denis-parnovskiy: For now I have just added roles I'm aware of. Definitely we will have much bigger list.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,170 @@
...
+    ]
+}
+
+# The bucket policies are auto-generated from the bucket list with the names write-name, read-name. E.g. write_blobs
+
+# Roles. The roles are the collection of the policies.
+roles = {
+    # E.g. nucleus needs this role
+    "write_blobs_read_logs": [
+        "write_blobs",
```
ddl-denis-parnovskiy, line 61: The change from previous revision:
* role name is changed from "nucleus" to the name reflecting the contents of the role
* instead of crating one policy write_blobs_read_logs, we now create 2 separate policies read_blobs and write_blobs and add them to the role.

cdk/domino_cdk/provisioners/s3.py
```diff
@@ -72,23 +61,5 @@ def provision_buckets(self, name: str, s3: List[config.S3]):
...
-        self.policy = iam.ManagedPolicy(
-            self.scope,
-            "S3",
-            managed_policy_name=f"{name}-S3",
-            statements=[
-                iam.PolicyStatement(
-                    actions=[
-                        "s3:ListBucket",
-                        "s3:GetBucketLocation",
-                        "s3:ListBucketMultipartUploads",
```
Secretions, line 88: `ListBucket` isn't present in the new permissions at all, and the other two here are now gated to the bucket. I believe (though don't know for a fact) that these permissions were here and ungated for a reason. That reason may be down to how Nucleus or some other service accesses the buckets. We can leave it how it is and see how things go, but at the very least thought I should mention it.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
```diff
@@ -0,0 +1,170 @@
...
+                f"{stack_name}-read-{name}",
+                managed_policy_name=f"{stack_name}-read-{name}",
+                statements=[
+                    iam.PolicyStatement(
+                        actions=s3_read_permissions,
+                        resources=[f"{bucket.bucket_arn}*"],
+                    )
+                ],
+            )
+            policies[f"write_{name}"] = iam.ManagedPolicy(
```
Secretions, line 159: I'm not sure what alternatives I have to offer for the configuration paradigm, but I'd prefer not to rely on composed strings for purely internally accessed things if possible (ie as opposed to something like `policies["blobs"]["write"]`).

cdk/domino_cdk/provisioners/eks/eks_iam.py
```diff
@@ -11,7 +12,7 @@ def __init__(
     ) -> None:
         self.scope = scope
 
-    def provision(self, name: str, cluster_name: str, s3_policy: iam.ManagedPolicy, r53_zone_ids: List[str]):
+    def provision(self, stack_name: str, cluster_name: str, r53_zone_ids: List[str], buckets: Dict[str, Bucket]):
```
Secretions, line 15: Regardless of whether we prefer `name` or `stack_name`, I don't think we should have it as `stack_name` in just these two files and still `name` everywhere else.

cdk/domino_cdk/provisioners/eks/eks_iam.py
```diff
@@ -40,56 +41,89 @@ def provision(self, name: str, cluster_name: str, s3_policy: iam.ManagedPolicy,
...
             autoscaler_policy,
             iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKSWorkerNodePolicy'),
             iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEC2ContainerRegistryReadOnly'),
             iam.ManagedPolicy.from_aws_managed_policy_name('AmazonEKS_CNI_Policy'),
             iam.ManagedPolicy.from_aws_managed_policy_name('AmazonSSMManagedInstanceCore'),
         ]
+
         if r53_zone_ids:
-            managed_policies.append(route53_policy)
+            managed_policies.append(
```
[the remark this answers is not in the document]
ddl-denis-parnovskiy: Yep. I was thinking about that but it is the old code. I'm just merging 2 separate "if r53:"

This function already should fail flake8 test for "too complex"

cdk/domino_cdk/provisioners/eks/eks_iam.py
ddl-denis-parnovskiy: I really prefer "stack_name" everywhere. Just "name" is too generic and already caused a confusion in the previous iteration of this PR.

Do you ask to go through the code and replace name to stack_name?

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
ddl-denis-parnovskiy: I'm fine with this.

cdk/domino_cdk/provisioners/s3.py
ddl-denis-parnovskiy: Yes, I was copying these permission from somewhere else, not from the node permissions.

ListBucket cannot be bound to a single bucket because it lists al buckets.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
ddl-denis-parnovskiy: Why it is done this way:

We want to have internal config with the roles that refer to the policies.

Best way to refer is by some reference ID, in this case, policy name.

If we put policies in some kind of language-based memory tree structure, we cannot refer them by the ID anymore.

Furthermore, since now I generate the bucket-based policies in runtime, based on the bucket list from the config, now I cannot have role definitions before the policies are programmatically generated.

An alternative to this is to predefine the whole policy tree structure before the provisioner is called. But then it is redundant and represents an anti-pattern of "second source of truth" for the bucket list. Then the bucket list will come both from the config and from internal roles config.

TBH I don't really understand your resistance to the concatenated strings. 
Even is simple hash map, if you have a key (a, b) , then internally if will be concatenated and become hash(a.concat(b)) anyway.
Creating of two level tree structure to just hold a list of the policies we want to refer in a human-readable way...

I could make a tree structure and refer the policies as some kind of 2-part formatted string like "blobs/read" and then parse the reference to split it into 2 parts to use as 2 indexes. But I think it is overkill to just to avoid the string concatenation.

cdk/domino_cdk/provisioners/eks/eks_iam.py
Secretions: It could be in a followup PR if you prefer, but I'd like to keep it consistent across all the provisioners.

ddl-denis-parnovskiy: If it just in provisioners, I'll do this.

cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py
ddl-denis-parnovskiy: Actually, not a big change.

Would you agree on:
1. Put the policies in 2-level tree like s3_policies["blobs'']["read"]
2. Refer them as "blobs-read" in the flat policy list and then just to split("-")[0]

?

ddl-denis-parnovskiy: Ah. I still need to concatenate the string to form the external policy name.

I.e. if we have stack denis-test and ["blobs"]["read"] we still create a policy "denis-test-blobs-read". I.e. stack_name + bucket_name + mode.

Would in invalidate your "for internal purposes only" objection?

ddl-denis-parnovskiy: Ah. I still need to concatenate the string to form the external policy name.

I.e. if we have stack denis-test and ["blobs"]["read"] we still create a policy "denis-test-blobs-read". I.e. stack_name + bucket_name + mode.

Would in invalidate your "for internal purposes only" objection?

Secretions: I did say I'm not sure what alternatives I have to offer. :)

Something like the two level tree is what I was thinking, but still avoiding string manipulation (so do the same thing on the list side).

Here's a quick (completely untested) attempt: https://gist.github.com/Secretions/5d925ac0a876a7b4ee00b9ca49db5efe

(I might also reverse the order in the policy names if I did this approach, to match the structure, ie name-read, name-write)

ddl-denis-parnovskiy: Yay-yay. Wait a min

ddl-denis-parnovskiy: Undertested, but take a look

ddl-denis-parnovskiy: Names:

```
┌───┬───────────────────────────────────────────────┬────────┬───────────────────────────────┬─────────────────────────────────────────────────┬────────────────────────────────────────────┐
│   │ Resource                                      │ Effect │ Action                        │ Principal                                       │ Condition                                  │
├───┼───────────────────────────────────────────────┼────────┼───────────────────────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ + │ ${denis-test1-4SA-blobs-write.Arn}            │ Allow  │ sts:AssumeRole                │ Service:eks.amazonaws.com                       │                                            │
│ + │ ${denis-test1-4SA-blobs-write.Arn}            │ Allow  │ sts:AssumeRoleWithWebIdentity │ Federated:${eks/OpenIdConnectProvider/Resource} │ "StringLike": "${OidcJson/Resource.Value}" │
├───┼───────────────────────────────────────────────┼────────┼───────────────────────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ + │ ${denis-test1-4SA-blobs-write--logs-read.Arn} │ Allow  │ sts:AssumeRole                │ Service:eks.amazonaws.com                       │                                            │
│ + │ ${denis-test1-4SA-blobs-write--logs-read.Arn} │ Allow  │ sts:AssumeRoleWithWebIdentity │ Federated:${eks/OpenIdConnectProvider/Resource} │ "StringLike": "${OidcJson/Resource.Value}" │
├───┼───────────────────────────────────────────────┼────────┼───────────────────────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ + │ ${denis-test1-4SA-images-write.Arn}           │ Allow  │ sts:AssumeRole                │ Service:eks.amazonaws.com                       │                                            │
│ + │ ${denis-test1-4SA-images-write.Arn}           │ Allow  │ sts:AssumeRoleWithWebIdentity │ Federated:${eks/OpenIdConnectProvider/Resource} │ "StringLike": "${OidcJson/Resource.Value}" │
└───┴───────────────────────────────────────────────┴────────┴───────────────────────────────┴─────────────────────────────────────────────────┴────────────────────────────────────────────┘
IAM Policy Changes
┌───┬───────────────────────────────────────────┬───────────────────────────────┐
│   │ Resource                                  │ Managed Policy ARN            │
├───┼───────────────────────────────────────────┼───────────────────────────────┤
│ + │ ${denis-test1-4SA-blobs-write}            │ ${denis-test1-S3-blobs-write} │
├───┼───────────────────────────────────────────┼───────────────────────────────┤
│ + │ ${denis-test1-4SA-blobs-write--logs-read} │ ${denis-test1-S3-blobs-write} │
│ + │ ${denis-test1-4SA-blobs-write--logs-read} │ ${denis-test1-S3-logs-read}   │
├───┼───────────────────────────────────────────┼───────────────────────────────┤
│ + │ ${denis-test1-4SA-images-write}           │ ${denis-test1-ECR-ecr-write}  │
└───┴───────────────────────────────────────────┴───────────────────────────────┘
```

Secretions: What is your opinion on this particular difference between the gist and the commit you pushed?

```
    "write_blobs_read_logs": [
        {"blobs": "write"},
        {"logs": "read"},
    ],
[...]
            for svc, policy_id in policy_list:
                iam_role.add_managed_policy(managed_policies[svc][policy_id])
```

It avoids the string split, and the policy list cleanly maps onto the managed_policy's data structure.

ddl-denis-parnovskiy: Agrh. I really hate forcing config writers to write extra { "", "" } where they can write in semi-plaintext ":".

I.e. yaml beats python :)

OK, if it will make you agree, I can do this. Not a huge difference

Secretions: Thanks, sorry for the trouble. :)

ddl-denis-parnovskiy: No worries. It is fun.

ddl-denis-parnovskiy: This way: see diff

Secretions: Looks good

Secretions (approved):

PR #27, opened 2021-06-20
Adding secrets encryption with KMS key (PLAT-2263)

ddl-denis-parnovskiy:
The change is straightforward. Add config field, use in in eks.Cluster.

### DIFF
--- a/cdk/domino_cdk/config/__init__.py
+++ b/cdk/domino_cdk/config/__init__.py
@@ -1,4 +1,5 @@
 from semantic_version import Version
+from typing import Optional
 
 from domino_cdk import __version__
 from domino_cdk.config.base import DominoCDKConfig
@@ -21,6 +22,7 @@ def config_loader(c: dict):
 def config_template(
     bastion: bool = False,
     private_api: bool = False,
+    secrets_encryption_key_arn: Optional[str] = None,
     dev_defaults: bool = False,
 ):
     max_nodegroup_azs = 3
@@ -69,6 +71,7 @@ def config_template(
             "eks": {
                 "version": "1.19",
                 "private_api": private_api,
+                "secrets_encryption_key_arn": secrets_encryption_key_arn,
                 "max_nodegroup_azs": max_nodegroup_azs,
                 "global_node_labels": {"dominodatalab.com/domino-node": "true"},
                 "global_node_tags": {},
--- a/cdk/domino_cdk/config/eks.py
+++ b/cdk/domino_cdk/config/eks.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from domino_cdk.config.util import MachineImage, check_leavins, from_loader
 
@@ -13,6 +13,7 @@ class EKS:
     max_nodegroup_azs: 3 - Will provision nodegroups in up to this many availability zones.
     global_node_labels: some-label: "true"  - Labels to apply to all kubernetes nodes
     global_node_tags: some-tags: "true"  - Labels to apply to all kubernetes nodes
+    secrets_encryption_key_arn: ARN  - KMS key arn to encrypt kubernetes secrets. A new key will be created if omitted.
     """
 
     @dataclass
@@ -107,6 +108,7 @@ class EKS:
     global_node_tags: Dict[str, str]
     managed_nodegroups: Dict[str, ManagedNodegroup]
     unmanaged_nodegroups: Dict[str, UnmanagedNodegroup]
+    secrets_encryption_key_arn: Optional[str]
 
     def __post_init__(self):
         errors = []
@@ -152,6 +154,7 @@ class EKS:
                 unmanaged_nodegroups={
                     name: EKS.UnmanagedNodegroup.load(ng) for name, ng in c.pop("nodegroups", {}).items()
                 },
+                secrets_encryption_key_arn=None,
             ),
             c,
         )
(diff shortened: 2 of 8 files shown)


ddl-denis-parnovskiy:
Discussion: should we make it optional?

Background: rule 5.3.1 requires "Ensure Kubernetes Secrets are encrypted using Customer Master Keys (CMKs) managed in AWS KMS". 

On the other hand, using default AWS keys is considered "Meduim security risk"

On the third hand I cannot find any "default AWS keys" to use for envelope encryption. 

By default the keys are encrypted at rest.

cdk/domino_cdk/config_template.yaml
```diff
@@ -40,6 +40,7 @@ route53:
 eks:
   version: "1.19"
   private_api: false
+  secrets_encryption_key_arn: arn:aws:kms:us-west-2:890728157128:key/b9c778f3-46cc-4416-8867-4ab47d912bf6
```
ddl-denis-parnovskiy, line 43: This is one of our keys in AWS KMS. If we go forward with making this field mandatory, we need to put some key ARN by default. There is no difference which key to use.

Secretions:
Regarding the optionality: I think we should either make it optional, or make it so there is a version of this where we provision a KMS for the cluster for when we're not providing a pre-existing one.

Also, we don't have CI setup in this repo yet, but we do have unit tests for the configs. Would you mind validating that this (or the final version of this) passes them?

ddl-denis-parnovskiy:
OK. I'll make it optional but add a warning message.

cdk/domino_cdk/config/__init__.py
```diff
@@ -69,6 +70,7 @@ def config_template(
             "eks": {
                 "version": "1.19",
                 "private_api": private_api,
+                "secret_encrption_key_name": secrets_encryption_key_arn,
```
steved, line 73:
```suggestion
                "secrets_encryption_key_arn": secrets_encryption_key_arn,
```

cdk/domino_cdk/config/eks.py
```diff
@@ -107,6 +111,7 @@ def load(cls, ng):
     global_node_tags: Dict[str, str]
     managed_nodegroups: Dict[str, ManagedNodegroup]
     unmanaged_nodegroups: Dict[str, UnmanagedNodegroup]
+    secrets_encryption_key_arn: str
```
steved, line 114: optional?

cdk/domino_cdk/provisioners/eks/eks_cluster.py
```diff
@@ -37,6 +39,16 @@ def provision(
...
+            secrets_encryption_key=Key.from_key_arn(
+                self.scope, "secrets_encryption_key_arn", secrets_encryption_key_arn
+            )
+            if secrets_encryption_key_arn
+            else Key(
+                self.scope,
+                f"{name}-kubernetes-secrets-envelope-key",
+                alias=f"{name}-kubernetes-secrets-envelope-key",
+                removal_policy=cdk.RemovalPolicy.DESTROY,
+            ),
```
steved, line 51: Can we also add `enableKeyRotation=true`?

cdk/domino_cdk/config/__init__.py
ddl-denis-parnovskiy: Let's not be confusing. It is ARN, not a name.

cdk/domino_cdk/config/eks.py
ddl-denis-parnovskiy: Ah, type.

cdk/domino_cdk/config/__init__.py
steved: Sorry! I just meant to highlight the type in `secret_encryption_key_name`, totally miffed the copy-paste.

ddl-denis-parnovskiy: Thanks!

cdk/domino_cdk/provisioners/eks/eks_cluster.py
ddl-denis-parnovskiy: Need to try. The doc doesn't say we cannot.

ddl-denis-parnovskiy: This works at least during deployment.

steved: Thanks! It comes in the CIS compliance scans.

steved (approved):
LGTM, though maybe @Secretions should have final say.

cdk/domino_cdk/config/eks.py
```diff
@@ -1,8 +1,11 @@
 from dataclasses import dataclass
-from typing import Dict, List
+from typing import Dict, List, Optional
+from logging import Logger
 
 from domino_cdk.config.util import MachineImage, check_leavins, from_loader
 
+log = Logger("domino_cdk.config")
```
Secretions, line 7: We don't actually use this and its corresponding import (was for an earlier version of the PR).

cdk/domino_cdk/provisioners/eks/eks_cluster.py
Secretions: Just a personal preference thing, but this ternary block is big enough that I'd be inclined to evaluate it before the call to `eks.Cluster()`

cdk/domino_cdk/config/eks.py
```diff
@@ -13,6 +16,7 @@ class EKS:
     max_nodegroup_azs: 3 - Will provision nodegroups in up to this many availability zones.
     global_node_labels: some-label: "true"  - Labels to apply to all kubernetes nodes
     global_node_tags: some-tags: "true"  - Labels to apply to all kubernetes nodes
+    secrets_encryption_key_arn: None  - KMS key name/arn to encrypt kubernetes secrets
```
Secretions, line 19: Should mention that if this is `None` we'll create the key, as this comment will get added to the generated config file (which will be replacing the static `config_template.yaml`).

Secretions (approved):

PR #28, opened 2021-06-21
Patching cluster config to enable logging (PLAT-2262)

ddl-denis-parnovskiy:
Use AwsCustomResource to execute update function

### DIFF
--- a/cdk/domino_cdk/provisioners/eks/eks_cluster.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_cluster.py
@@ -1,5 +1,7 @@
 import aws_cdk.aws_ec2 as ec2
 import aws_cdk.aws_eks as eks
+import aws_cdk.aws_logs as logs
+import aws_cdk.custom_resources as cr
 from aws_cdk import core as cdk
 from aws_cdk.aws_kms import Key
 
@@ -39,7 +41,9 @@ class DominoEksClusterProvisioner:
                 enable_key_rotation=True,
             )
 
-        # Note: We can't tag the EKS cluster via CDK/CF: https://github.com/aws/aws-cdk/issues/4995
+        # TODO: tag the EKS cluster
+        # Need some way: https://github.com/aws/aws-cdk/issues/4995
+        
         cluster = eks.Cluster(
             self.scope,
             "eks",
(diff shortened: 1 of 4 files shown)


Secretions (commented):
Any chance we could piggyback on this and also have it tag the eks cluster?

We're not tagging it because there is no cloudformation api for tagging eks cluster (see: https://github.com/aws/aws-cdk/issues/4995), so one of the things I wanted to create a big provisioning lambda to do is work around this lack of tagging.

ddl-denis-parnovskiy:
> Any chance we could piggyback on this and also have it tag the eks cluster?

We cannot just add tags to this function call. 

But there is a good chance to use the same approach to call tagging API. 

AwsCustomResource is converted into JS call.

JS SDK syntax is as follows:
```
var params = {
  ResourceARNList: [ /* required */
    'STRING_VALUE',
    /* more items */
  ],
  Tags: { /* required */
    '<TagKey>': 'STRING_VALUE',
    /* '<TagKey>': ... */
  }
};
resourcegroupstaggingapi.tagResources(params, function(err, data) {
  if (err) console.log(err, err.stack); // an error occurred
  else     console.log(data);           // successful response
});
```

What I do not know here is the relation between "resourcegroupstaggingapi" and `service` field. An experiment could answer this. Or not :)

I can take it.

ddl-denis-parnovskiy:
Changing to WIP because I want to add "on_delete" call to delete the logging group on cluster tear down. Now it is not deleted automatically because it is not part of CF stack.

Secretions (approved):

cdk/domino_cdk/provisioners/eks/eks_cluster.py
```diff
@@ -53,6 +57,33 @@ def provision(
...
+            "UpdateClusterConfigCustom",
+            timeout=cdk.Duration.minutes(10),  # defaults to 2 minutes
+            log_retention=logs.RetentionDays.ONE_DAY,  # defaults to never delete logs
+            policy=cr.AwsCustomResourcePolicy.from_sdk_calls(resources=cr.AwsCustomResourcePolicy.ANY_RESOURCE),
+            on_create={
+                "service": "EKS",
+                "action": "updateClusterConfig",
+                "parameters": params,
+                "physical_resource_id": cr.PhysicalResourceId.of("UpdateClusterConfigCustom"),
+                "ignore_error_codes_matching": "...",
```
Secretions, line 83: A comment explaining why this is here might not be a bad idea.

PR #31, opened 2021-06-25
PLAT-2421: S3 improvements and general compliance / security changes

steved:
* S3 improvements: access logging, versioning enabled by default
* Optional VPC flow logging
* ELB access logging by default (opt-out by setting the monitoring bucket to null)
* Switches bastion to AL2 for consistency's sake
* Encryption by default for EBS / EFS
* Switch back to EFS burst / standard tier as this is our current default (up for debate though)
* Disable public IP mapping for public subnets (we use EIPs and ELBs do their own mapping)
* Remove all rules from the default SG
* Support govcloud partition for IAM rules
* Default K8s version is 1.20


cdk/domino_cdk/provisioners/s3.py
```diff
@@ -25,55 +31,120 @@ def __init__(self, parent: cdk.Construct, construct_id: str, name: str, s3: List
...
+        bucket_id: str,
+        attrs: config.S3.Bucket,
+        server_access_logs_bucket: Optional[Bucket] = None,
+        **kwargs,
+    ) -> Bucket:
+        use_sse_kms_key = False
+        if attrs.sse_kms_key_id:
+            use_sse_kms_key = True
+            # TODO: is this correct???
+            sse_kms_key = Key.from_key_arn(self.scope, f"{bucket_id}-kms-key", attrs.sse_kms_key_id)
```
steved, line 46: this line used to be `Key.from_key_arn(self, f"{bucket}-kms-key", attrs.sse_kms_key_id)`, which had mypy complaining...

cdk/domino_cdk/agent.py
```diff
@@ -27,6 +27,11 @@ def generate_install_config(
                 }
             },
         },
+        "namespaces": {
+            "platform": {"name": f"{name}-platform"},
+            "compute": {"name": f"{name}-compute"},
+            "system": {"name": f"{name}-system"},
```
Secretions, line 33: I ended up omitting these intentionally and just going with the default `domino-xxx`. Seemed like in the non-deployer world nobody ever cared to customize it, and we're really only scripting (and setup to do) single-Domino installs anyway, so I figured why not just do the simple, uniform thing?

cdk/domino_cdk/config/s3.py
```diff
@@ -13,30 +13,37 @@ class S3:
     removal_policy_destroy: true/false - Delete bucket when destroying stack. If auto_delete_objects is false,
                                          destroys will fail unless buckets are emptied.
     sse_kms_key_id: XXX - Specific KMS key ID to setup with bucket. Otherwise, defaults to AES256
+    require_encryption: true/false - Require all requests and objects to be encrypted through policies.
+                                     Defaults to true, changing to false does not disable default KMS key encryption of objects.
```
Secretions, line 17: Should we maybe just leave this hard on?

cdk/domino_cdk/provisioners/efs.py
```diff
@@ -34,15 +34,17 @@ def provision_efs(self, name: str, cfg: config.EFS, vpc: ec2.Vpc, security_group
...
             vpc=vpc,
-            # encrypted=True,
+            encrypted=True,
             file_system_name=name,
             # kms_key,
             # lifecycle_policy,
-            performance_mode=efs.PerformanceMode.MAX_IO,
-            provisioned_throughput_per_second=cdk.Size.mebibytes(100),  # TODO: dev/nondev sizing
+            # performance_mode=efs.PerformanceMode.MAX_IO,
+            performance_mode=efs.PerformanceMode.GENERAL_PURPOSE,
```
Secretions, line 42: Should we make these performance things configurable? Also what about kms?

cdk/domino_cdk/provisioners/eks/__init__.py
```diff
@@ -41,8 +41,10 @@ def __init__(
         )
 
         cdk.CfnOutput(parent, "eks_cluster_name", value=self.cluster.cluster_name)
+
+        region = cdk.Stack.of(self.scope).region
```
Secretions, line 45: What is this doing differently?

cdk/domino_cdk/provisioners/eks/eks_cluster.py
```diff
@@ -2,6 +2,7 @@
 
 import aws_cdk.aws_ec2 as ec2
 import aws_cdk.aws_eks as eks
+import aws_cdk.aws_iam as iam
```
Secretions, line 5: Presume this was left behind from a previous iteration?

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -88,23 +140,51 @@ def provision_vpc(self, name: str, vpc: config.VPC):
                 subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PRIVATE),
             )
 
-    def provision_bastion(self, name: str, bastion: config.VPC.Bastion) -> None:
+        # TODO until https://github.com/aws/aws-cdk/issues/14194
+        for idx, subnet_id in enumerate(self.vpc.select_subnets(subnet_type=ec2.SubnetType.PUBLIC).subnet_ids):
+            name = f"DisableMapPublicIpLaunch-{idx}"
+            cr.AwsCustomResource(
```
Secretions, line 146: I think we might be heading towards a place where we should just rip out the single-stack option.

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -88,23 +140,51 @@ def provision_vpc(self, name: str, vpc: config.VPC):
...
         if not bastion.enabled:
             return None
         if bastion.ami_id:
             machine_image = ec2.MachineImage.generic_linux(
+                # TODO no region???
                 {self.region: bastion.ami_id},
                 user_data=ec2.UserData.custom(bastion.user_data),
             )
         else:
+            # ??? is this now covered under the config? is it not auto?? why is it in here?
```
Secretions, line 181: Ostensibly, you could synth a generic cloudformation template rather than a deployment-specific one by not providing a region or account id, and I tried to keep that an option early on. Was also going to have a few pre-fab cloudformation templates. For various reasons this really didn't work out, but the bastion support was the first thing that really hard broke the possibility of doing this.

So, conceptually, if you had bastion off, you could not provide the aws account id and still proceed.

I think it's fine to move this check to config and just make it a hard requirement.

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -88,23 +140,51 @@ def provision_vpc(self, name: str, vpc: config.VPC):
...
             if not self.scope.account.isnumeric():  # TODO: Can we get rid of this requirement?
                 raise ValueError(
-                    "Error loooking up AMI: Must provide explicit AWS account ID to do AMI lookup. Either provide AMI ID or AWS account id"
+                    "Error looking up AMI: Must provide explicit AWS account ID to do AMI lookup. Either provide AMI ID or AWS account id"
                 )
 
-            machine_image = ec2.LookupMachineImage(
-                name="ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-*", owners=["099720109477"]
-            )
+            machine_image = ec2.AmazonLinuxImage()
```
Secretions, line 187: Oh, and this will require a corresponding deployer PR due to the username (could just hardcode the username into the cdk template) fyi

Secretions: The iam region arg as well, actually.

steved: Ah yep, sorry, got a few things pending here: https://github.com/cerebrotech/platform-apps/pull/3866

cdk/domino_cdk/provisioners/eks/__init__.py
steved: mypy complains, but ultimately I think this is basically casting the "reusable" `Construct` into ` Stack` that has the actual region attribute. Question is whether or not we should be pedantically correct, I guess.

cdk/domino_cdk/agent.py
steved: :+1: will revert

cdk/domino_cdk/config/s3.py
steved: Yeah this shouldn't be exposed as a config option. It's required for the monitoring / access logs bucket, but it can be an internal flag only.

cdk/domino_cdk/config/template.py
```diff
@@ -127,7 +128,12 @@ def add_nodegroups(name, count, min_size, instance_types, labels, disk_size=disk
             'registry': S3.Bucket(
                 auto_delete_objects=destroy_on_destroy, removal_policy_destroy=destroy_on_destroy, sse_kms_key_id=None
             ),
-        }
+        },
+        monitoring_bucket=S3.Bucket(
```
ddl-denis-parnovskiy, line 132: I think you have a config option for this one. But it is not used here.

cdk/domino_cdk/domino_stack.py
```diff
@@ -31,7 +31,11 @@ def __init__(
             cdk.Tags.of(self).add(str(k), str(v))
 
         self.s3_stack = DominoS3Provisioner(self, "S3Stack", self.name, self.cfg.s3, nest)
-        self.vpc_stack = DominoVpcProvisioner(self, "VpcStack", self.name, self.cfg.vpc, nest)
+
+        self.vpc_stack = DominoVpcProvisioner(
+            self, "VpcStack", self.name, self.cfg.vpc, nest, flow_log_bucket=self.s3_stack.buckets.get("monitoring")
```
ddl-denis-parnovskiy, line 36: Would it fail if monitoring_bucket is False in the config?

cdk/domino_cdk/provisioners/efs.py
```diff
@@ -34,15 +34,14 @@ def provision_efs(self, stack_name: str, cfg: config.EFS, vpc: ec2.Vpc, security
             self.scope,
             "Efs",
             vpc=vpc,
-            # encrypted=True,
+            encrypted=True,
```
ddl-denis-parnovskiy, line 37: What are the costs? Should it be an option? Why it was commented out?

cdk/domino_cdk/config/template.py
steved: Explain a bit more, not sure I understand?

cdk/domino_cdk/domino_stack.py
steved: At the moment, no it checks within DominoVpcProvisioner and raises. I'll add a test case.

cdk/domino_cdk/provisioners/efs.py
steved: No cost associated, this is the default for existing EKS deploys. No sure why it was commented out tbh.

Secretions: I think it was to investigate this documentation comment that got cut off in the docs:

```
encrypted (Optional[bool]) – Defines if the data at rest in the file system is encrypted or not. Default: - If your application has the ‘
```

Looking at the code, it's this:

```
   * @default - If your application has the '@aws-cdk/aws-efs:defaultEncryptionAtRest' feature flag set, the default is true, otherwise, the default is false.
```

(so we do need it)

Also wondering, do we need to do anything with kms here?

cdk/domino_cdk/config/template.py
ddl-denis-parnovskiy: I probably do not understand how we deal with the optional parameters here. 

monitoring_bucket is optional but it is created here without any comment

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -88,23 +140,46 @@ def provision_vpc(self, stack_name: str, vpc: config.VPC):
...
+                on_create=cr.AwsSdkCall(
+                    action="modifySubnetAttribute",
+                    service="EC2",
+                    parameters={"MapPublicIpOnLaunch": {"Value": False}, "SubnetId": subnet_id},
+                    physical_resource_id=cr.PhysicalResourceId.of(name),
+                ),
+            )
+
+        if vpc.flow_logging:
+            if not flow_log_bucket:
```
ddl-denis-parnovskiy, line 160: are 
* monitoring
* monitoring_bucket
* mb
* flow_log_bucket

mean the same thing? Could we name them the same?

cdk/domino_cdk/config/template.py
steved: Ah yeah that's totally fair. I'll add actual documentation.

cdk/domino_cdk/provisioners/vpc.py
steved: Yep!

cdk/domino_cdk/config/template.py
```diff
@@ -21,17 +21,16 @@ def config_template(
     destroy_on_destroy = False
     disk_size = 1000
     platform_instance_type = "m5.2xlarge"
-    platform_min_size = 3
+
     if dev_defaults:
         max_nodegroup_azs = 1
         destroy_on_destroy = True
         disk_size = 100
```
ddl-denis-parnovskiy, line 28: do you still need this?

cdk/domino_cdk/config/template.py
```diff
@@ -52,16 +51,17 @@ def add_nodegroups(name, count, min_size, instance_types, labels, disk_size=disk
...
-        platform_min_size,
+        1,
         [platform_instance_type],
         {"dominodatalab.com/node-pool": "platform"},
     )
     add_nodegroups(
         "compute",
         compute_nodegroups,
-        1,
+        0,
```
ddl-denis-parnovskiy, line 61: I vaguely remember something like 0 is not a valid value

ddl-denis-parnovskiy (approved):

cdk/domino_cdk/config/template.py
ddl-denis-parnovskiy: In later lines you change disk_size=disk_size to disk_size=100

steved: Yes, I believe so? For "prod", we do want 1000G disks for compute & GPU and there isn't a way to set a variable optionally like that, no?

cdk/domino_cdk/config/template.py
steved: For managed groups, yeah, there's still some pending cluster-autoscaler support. For unmanaged, it seems to work fine.

cdk/domino_cdk/config/template.py
ddl-denis-parnovskiy: I mean you have outer var disk_size and inner was disk_size in add_nodegroups. And you pass one into another. But add_nodegroups is already a closure that sees outer disk_size.

Secretions (approved):

cdk/domino_cdk/config/template.py
ddl-denis-parnovskiy: Sorry for confusion, I'm talking about the different line, the definition of add_nodegroups().
It sees disk_size. Do you want to pass it?

ddl-denis-parnovskiy: You have my approval anyway.

Just passing outer scope variables into the closure looks off to me

PR #35, opened 2021-07-01
Use lambdas for some tasks (PLAT-2427, PLAT-2428, PLAT-2429, PLAT-2469)

ddl-denis-parnovskiy:
A function to create lambdas.

A Lambda to tag cluster, enable k8s logging, change k8s logs retention to 7 days
* On delete this lambda deletes k8s logs

A Lambda to find and delete all recovery points of our backup

A lambda that is topologically above the cluster resource so its on_delete happens after the cluster is deleted, on delete it changes retention of all cluster log groups to 1 day, so the logs will automatically disappear next day.


cdk/domino_cdk/provisioners/efs.py
```diff
@@ -103,3 +107,19 @@ def provision_backup_vault(self, stack_name: str, efs_backup: config.EFS.Backup)
...
+
+        DominoCdkUtil.create_lambda(
+            scope=self.scope,
+            stack_name=stack_name,
+            dirname=path.dirname(path.abspath(__file__)),
+            name="backup_post_creation_tasks",
+            environment={"stack_name": stack_name, "backup_vault": vault.backup_vault_name},
+            resources=[
+                "arn:aws:logs:*",
+                f"arn:aws:backup:{self.scope.region}:{self.scope.account}:backup-vault/{stack_name}-efs*",
```
steved, line 119: Please make sure the partition uses the scope or use the `Arn` helper so this works in govcloud, e.g.
```
f"arn:{partition}:...
```
or 
https://github.com/dominodatalab/cdk-cf-eks/blob/89dc604e80e5768f034c576553ce21ad55179fd5/cdk/domino_cdk/provisioners/eks/eks_iam.py#L58

cdk/domino_cdk/provisioners/eks/eks_cluster.py
```diff
@@ -57,54 +77,29 @@ def provision(
...
-                "parameters": params,
-                "physical_resource_id": cr.PhysicalResourceId.of("TagClusterCustom"),
+        DominoCdkUtil.create_lambda(
+            scope=self.scope,
+            stack_name=stack_name,
+            dirname=path.dirname(path.abspath(__file__)),
+            name="cluster_post_creation_tasks",
+            environment={
+                "cluster": cluster.cluster_name,
+                "cluster_arn": f"arn:aws:eks:{self.scope.region}:{self.scope.account}:cluster/{cluster.cluster_name}",
```
steved, line 89: Does `cluster.cluster_arn` work here?

cdk/domino_cdk/provisioners/eks/lambda/cluster_post_creation_tasks.py
```diff
@@ -0,0 +1,56 @@
+import json
```
steved, line 1: Is it possible to unit test this code somehow? We've used [moto](https://github.com/spulec/moto) for AWS-mocking before.

cdk/domino_cdk/util.py
```diff
@@ -146,3 +153,60 @@ def overlay(alpha: dict, omega: dict) -> dict:
...
+                name + "_is_complete",
+                runtime=lambda_.Runtime.PYTHON_3_7,
+                handler="index.is_complete",
+                code=lambda_.InlineCode(is_complete_code_body),
+                # timeout defaults to 2 minutes
+                log_retention=logs.RetentionDays.ONE_DAY,  # defaults to never delete logs
+            )
+            is_complete.add_to_role_policy(statement)
+        except IOError:
+            pass
```
steved, line 202: What would IOError in this? A missing file?

cdk/domino_cdk/provisioners/efs.py
ddl-denis-parnovskiy: Will do

cdk/domino_cdk/provisioners/eks/eks_cluster.py
ddl-denis-parnovskiy: perhaps. I was not aware it exists.

cdk/domino_cdk/provisioners/eks/lambda/cluster_post_creation_tasks.py
ddl-denis-parnovskiy: The value of this UT is low. This function is basically 4 calls with the static parameters or parameters copied from env.

Let's talk

cdk/domino_cdk/util.py
ddl-denis-parnovskiy: I'm still living in Python < 3.3

cdk/domino_cdk/provisioners/efs.py
ddl-denis-parnovskiy: This too?

https://github.com/dominodatalab/cdk-cf-eks/blob/89dc604e80e5768f034c576553ce21ad55179fd5/cdk/domino_cdk/provisioners/eks/eks_iam_roles_for_k8s.py#L140

ddl-denis-parnovskiy:
I have fixed all except UT.

cdk/domino_cdk/provisioners/eks/lambda/cluster_post_creation_tasks.py
ddl-denis-parnovskiy: As discussed, our goal is to make a unit-test framework for this kind of code.

This particular lambda is simple and just passes parameters from env to API but we can use is a guinea pig.

We will do unit tests as separate JIRA https://dominodatalab.atlassian.net/browse/PLAT-2494

cdk/domino_cdk/util.py
```diff
@@ -146,3 +153,60 @@ def overlay(alpha: dict, omega: dict) -> dict:
...
+        statement = iam.PolicyStatement()
+        for r in resources:
+            statement.add_resources(r)
+        for a in actions:
+            statement.add_actions(a)
+        on_event.add_to_role_policy(statement)
+
+        is_complete = None
+        try:
+            with open(path.join(dirname, "lambda", name + "_is_complete.py"), encoding="utf-8") as fp:
```
Secretions, line 189: This is an honest question, not a leading one, but could this just be rolled into main lambda code so we don't have to look for two files? Like if you provide the same lambda for both things, and `is_complete` is a function in the code, will it just do the right thing? Also, do we anticipate having a use case for `is_complete` any time soon? ATM it looks like it's a code path that won't get exercised, but could imagine there's something coming up for it.

cdk/domino_cdk/util.py
```diff
@@ -146,3 +153,60 @@ def overlay(alpha: dict, omega: dict) -> dict:
...
+            on_event_code_body = fp.read()
+        on_event = lambda_.Function(
+            scope,
+            name + "_on_event",
+            function_name=f"{stack_name}-{name}",
+            runtime=lambda_.Runtime.PYTHON_3_7,
+            handler="index.on_event",
+            code=lambda_.InlineCode(on_event_code_body),
+            environment=environment,
+            timeout=cdk.Duration.seconds(600),  # default is 3 seconds
```
Secretions, line 177: Could wait until it comes up, but depending on what other things we use this for we might find ourselves wanting to configure the timeout.

cdk/domino_cdk/util.py
```diff
@@ -146,3 +153,60 @@ def overlay(alpha: dict, omega: dict) -> dict:
             return {}
         base_dict = check_type(dictionaries[0])
         return base_dict if len(dictionaries) == 1 else overlay(base_dict, cls.deep_merge(*dictionaries[1:]))
+
+    @staticmethod
+    def create_lambda(
```
Secretions, line 158: I feel like `provisioners/lambda/__init__.py` might be a better place for this? Everything else in here is non-provisioning-related helper functions (and mostly in service of `util.py` except for deep_merge), while this generates actual cdk lambda objects.

Secretions: Or if there was a separate-of-concerns thing, I could also see something like `domino_cdk/provisioners/lambda.py` for this and the lambda directory moved to `domino_cdk/lambda` making sense? It's not a big deal either way, just seems more categorically in line with the provisioners to me.

cdk/domino_cdk/util.py
```diff
@@ -146,3 +153,60 @@ def overlay(alpha: dict, omega: dict) -> dict:
...
+
+        provider = cr.Provider(
+            scope,
+            name + "_provider",
+            on_event_handler=on_event,
+            is_complete_handler=is_complete,
+            log_retention=logs.RetentionDays.ONE_DAY,  # default is INFINITE
+        )
+
+        return cdk.CustomResource(scope, name + "_custom", service_token=provider.service_token)
```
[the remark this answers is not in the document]
ddl-denis-parnovskiy: Done

cdk/domino_cdk/util.py
ddl-denis-parnovskiy: 1. No, Provider class requires a separate lambda resource for is_complete. The only way to share the code is to create a bucket resource for the code and use it in 2 lambdas but this is 3 resources instead of 2.
2. Is complete is used to avoid his 10 minutes timeout for lambda. This way you have 2 functions, each completes in 3 seconds and doesn't use your container engine with the corresponding costs.
3. I delete Provider and is_complete because we really do not care.

cdk/domino_cdk/util.py
ddl-denis-parnovskiy: I'll add a parameter "timeout" to create_lambda() when we find that we want to configure it.

Now all lambdas complete when they are complete and do not use the timeout.

cdk/domino_cdk/util.py
ddl-denis-parnovskiy: I'm OK to move the function anywhere except __init__ of lambdas directory. lambdas directory contain files we upload to AWS. They are not a module, I'll remove accidental __init__ file to avoid confusion. 

The code to create lambda doesn't logically belong with the lambda bodies.

I'm OK to have one directory to hold all lambda bodies instead of having subdirectory along with every file using them.

Secretions (approved):

PR #36, opened 2021-07-07
PLAT-2013: Extra security group access controls

Secretions:


ddl-denis-parnovskiy:
Do we have a use case when they run something in compute that accepts API calls?

If this is the case, do we expect same personas to access nucleus and API running in the compute?

I probably say no. For example, the release engineer enables some WS run via nucleus and then there is a bot that feeds data to the WS using that API. E.g. In this case we do not want to give bot an access to the platform.

So my uneducated conclusion would be: we need separate access rules for platform and compute.

Secretions:
Will pick this back up when I get back.

PR #38, opened 2021-07-08
Save config as stack output + use ruamel for agent template output

Secretions:
* Running config is added to the stack as a CloudFormation output
  * This is just a safety / tracking / visibility thing. ie you can recover your settings if you lose your local copy, or if you're looking at what's provisioned you can know what the settings were, etc.
* The agent template now goes through ruamel's dump instead of the standard yaml one
  * Though the motivation was just not to have two different yaml dumpers in the same file, it has the advantage of preserving the order of the original structure, which mimics the order in the actual agent template

### DIFF
--- a/cdk/domino_cdk/util.py
+++ b/cdk/domino_cdk/util.py
@@ -1,5 +1,6 @@
 from filecmp import cmp
 from glob import glob
+from io import BytesIO
 from json import loads as json_loads
 from os.path import basename, isfile
 from os.path import join as path_join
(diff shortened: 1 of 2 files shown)


cdk/domino_cdk/util.py
```diff
@@ -146,3 +149,9 @@ def overlay(alpha: dict, omega: dict) -> dict:
             return {}
         base_dict = check_type(dictionaries[0])
         return base_dict if len(dictionaries) == 1 else overlay(base_dict, cls.deep_merge(*dictionaries[1:]))
+
+    @staticmethod
+    def ruamel_dump(data: dict):
+        buf = BytesIO()
+        YAML().dump(data, buf)
```
Secretions, line 156: Ruamel is, uhh, "eccentric" about not dumping to anything but a stream. Hence, wrapper.

ddl-denis-parnovskiy (approved):

PR #43, opened 2021-07-19
Upload release artifacts to S3 during CI

Secretions:
This PR adds the release artifacts created via `make dist` to an s3 bucket as publicly-accessible downloads.

Our current process for testing deployer integration involves creating a mutable pre-release (rc) tag in github, and iterating over it until the deployer PR is merged (at which point it becomes immutable). This is rather asinine.

With this, one can just use the s3 links CI uses to test, and then create a proper github release (still manually for now) once it's actually merged. No more iterating via github.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -60,8 +60,9 @@ jobs:
         AWS_ACCESS_KEY_ID: ${{ secrets.DOMINO_ARTIFACTS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.DOMINO_ARTIFACTS_ACCESS_KEY }}
         DOMINO_CDK_VERSION: ${{ github.sha }}
+        DATEDIR: "date +%Y%m%d"
       run: |
         cd ..
         make clean && make dist
-        aws s3 cp --acl=public-read ./dist/domino-cdk-$DOMINO_CDK_VERSION.tar.gz s3://domino-artifacts/cdk/
-        aws s3 cp --acl=public-read ./dist/domino-cdk-terraform-$DOMINO_CDK_VERSION.tar.gz s3://domino-artifacts/cdk/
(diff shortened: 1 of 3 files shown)


Secretions:
image
I also added a lifecycle rule to the bucket to delete uploads older than 45 days.

.github/workflows/python-app.yml
```diff
@@ -54,3 +55,17 @@ jobs:
       run: |
         cdk synth -q
         cdk synth --context nest=true -q
+    - name: Upload distribution artifacts
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.DOMINO_ARTIFACTS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.DOMINO_ARTIFACTS_ACCESS_KEY }}
+        DOMINO_CDK_VERSION: ${{ github.sha }}
+        DATEDIR: "date +%Y%m%d"
```
ddl-denis-parnovskiy, line 63: So, two developers/pipeline cannot use this on the same date? Both origins will use the same dir in S3?

Also, 23:59 it turns into a pumpkin and starts using a new dir?

Can we make it more local per pipeline/dev-env? Like: generate a random number and write it somewhere then use it?

Secretions: Everybody can use it on the same date as the artifacts themselves have the unique github sha for a given commit in the filename. There would have to be a hash collision with the sha for there to be a conflict. The date is only in the directory structure for convenience (so there aren't a zillion files in `domino-artifacts/cdk`'s root, and so it's easy to identify from roughly when an artifact was generated).

As an example, here are the terraform artifacts for two separate commits (the second is from draft PR #42, which also has a version of this code in it):

```
https://domino-artifacts.s3.amazonaws.com/cdk/20210719/domino-cdk-terraform-ca6a8f2c811b4623771f95dc4eb51f29affe44ba.tar.gz
https://domino-artifacts.s3.amazonaws.com/cdk/20210719/domino-cdk-terraform-7142ee95225f7d5c4a0f4d68fa960efa6eaa486d.tar.gz
```

ddl-denis-parnovskiy (approved):

PR #45, opened 2021-07-20
add SSM support & IMDSv2 requirement to bastion

steved:
Although disabled is safest, that doesn't pass AWS' "security compliance" check: 
https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#fsbp-ec2-8

In addition, although it seems a bit strange, we can allow the bastion to be configured with SSM if for some reason SSH doesn't work?

### DIFF
--- a/cdk/domino_cdk/provisioners/vpc.py
+++ b/cdk/domino_cdk/provisioners/vpc.py
@@ -225,6 +225,10 @@ class DominoVpcProvisioner:
             ),
         )
 
+        bastion_instance.role.add_managed_policy(
+            iam.ManagedPolicy.from_aws_managed_policy_name("AmazonSSMManagedInstanceCore"),
+        )
+
         ec2.CfnEIP(
             self.scope,
             "bastion_eip",
@@ -233,7 +237,7 @@ class DominoVpcProvisioner:
 
         cr.AwsCustomResource(
             self.scope,
-            "DisableBastionHTTPEndpoint",
+            "RequireBastionHTTPTokens",
             log_retention=logs.RetentionDays.ONE_DAY,
             policy=cr.AwsCustomResourcePolicy.from_sdk_calls(resources=cr.AwsCustomResourcePolicy.ANY_RESOURCE),
             on_update=cr.AwsSdkCall(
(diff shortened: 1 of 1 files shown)


ddl-denis-parnovskiy (approved):

ddl-denis-parnovskiy:
> we can allow the bastion to be configured with SSM if for some reason SSH doesn't work?

Is it a question?

steved:
Somewhat. Is it worth it, is it a security issue? The AMI has to support SSM.

Secretions (commented):
I feel like disabling the http endpoint is a better default security stance for when a bastion would be used normally? Maybe adding an SSM option toggle to the bastion is the right thing?

I sort of would have expected that an org that required this would also be one where the bastion would be off.

ddl-denis-parnovskiy:
Agree, we better have a flag.

steved:
Turns out IMDS is actually required for cloud-init to correctly configure the authorized SSH keys:
* The default datasource comes from IMDS: https://cloudinit.readthedocs.io/en/latest/topics/datasources/ec2.html
* which then reads the public key:
```
2021-07-12 20:42:37,782 - url_helper.py[DEBUG]: [0/6] open 'http://169.254.169.254/2016-09-02/meta-data/public-keys/' with {'url': 'http://169.254.169.254/2016-09-02/meta-data/public-keys/', 'allow_redirects': True, 'method': 'GET', 'timeout': 5.0, 'headers': {'X-aws-ec2-metadata-token': 'REDACTED'}} configuration
2021-07-12 20:42:37,786 - url_helper.py[DEBUG]: Read from http://169.254.169.254/2016-09-02/meta-data/public-keys/ (200, 13b) after 1 attempts
2021-07-12 20:42:37,786 - url_helper.py[DEBUG]: [0/6] open 'http://169.254.169.254/2016-09-02/meta-data/public-keys/0/openssh-key' with {'url': 'http://169.254.169.254/2016-09-02/meta-data/public-keys/0/openssh-key', 'allow_redirects': True, 'method': 'GET', 'timeout': 5.0, 'headers': {'X-aws-ec2-metadata-token': 'REDACTED'}} configuration
2021-07-12 20:42:37,790 - url_helper.py[DEBUG]: Read from http://169.254.169.254/2016-09-02/meta-data/public-keys/0/openssh-key (200, 393b) after 1 attempts
2021-07-12 20:42:37,791 - url_helper.py[DEBUG]: [0/6] open 'http://169.254.169.254/2016-09-02/meta-data/services/' with {'url': 'http://169.254.169.254/2016-09-02/meta-data
```
* which then configures the ec2-users' authorized_keys

It just happened by chance that since this was a separate API call, cloud-init could finish before the API call is done.

steved:
There's still the potential for race conditions if cloud-init chooses IMDSv1 for some reason, but going to merge this as its currently completely broken.

PR #46, opened 2021-07-20
support custom root device name for AMIs

steved:
We previously hardcoded `/dev/xvda`, but that only works for AMIs where the root device name matches. If it doesn't, we end up adding a new volume with that name and can potentially not get the root device volume size we need.

This solution is definitely a bit of a PITA for a potential corner case, but it is more "correct". The only speculative issue is the root volume size, which we can't obtain programmatically, because there really isn't a good way to reference it by value.

### DIFF
--- a/cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
@@ -262,9 +262,10 @@ class DominoEksNodegroupProvisioner:
         return mime_user_data
 
     def _launch_template(self, scope, name: str, ng: NodeGroup, **opts) -> ec2.LaunchTemplate:
-        root_device_name = "/dev/xvda"  # This only works for AL2
         if ng.ami_id:
             root_device_name = root_device_mapping(self.scope, ng.ami_id).name
+        else:
+            root_device_name = "/dev/xvda"  # This only works for AL2
 
         opts = {
             "key_name": ng.key_name,
--- a/cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional, TypeVar, Union
+from typing import Any, Dict, List, Optional, Union
 
 import aws_cdk.aws_ec2 as ec2
 import aws_cdk.aws_eks as eks
@@ -9,8 +9,6 @@ from aws_cdk import core as cdk
 from domino_cdk import config
 
 from ..ami import root_device_mapping
-
-NodeGroup = TypeVar("NodeGroup", bound=config.EKS.NodegroupBase)
 
 
 class DominoEksNodegroupProvisioner:
@@ -38,7 +36,7 @@ class DominoEksNodegroupProvisioner:
 
         max_nodegroup_azs = self.eks_cfg.max_nodegroup_azs
 
-        def provision_nodegroup(nodegroup: Dict[str, NodeGroup], prov_func):
+        def provision_nodegroup(nodegroup: Dict[str, config.eks.T_NodegroupBase], prov_func):
             for name, ng in nodegroup.items():
                 if not ng.ami_id:
                     ng.labels = {**ng.labels, **self.eks_cfg.global_node_labels}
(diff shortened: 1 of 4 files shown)


cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
```diff
@@ -281,3 +260,31 @@ def _handle_user_data(
                 mime_user_data.add_part(ec2.MultipartBody.from_user_data(ud))
 
         return mime_user_data
+
+    def _launch_template(self, scope, name: str, ng: NodeGroup, **opts) -> ec2.LaunchTemplate:
+        root_device_name = "/dev/xvda"  # This only works for AL2
```
ddl-denis-parnovskiy, line 265: root_device_name = "/dev/xvda"  # This only works for AL2 if not ng.ami_id else root_device_mapping(self.scope, ng.ami_id).name

I really don't like:
```
do-stuff
oh, wait, if some other stuff, then undo and redo stuff.
```
patterns

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -170,12 +173,16 @@ def provision_vpc(self, stack_name: str, vpc: config.VPC, monitoring_bucket: Opt
     def provision_bastion(self, name: str, bastion: config.VPC.Bastion) -> Optional[ec2.SecurityGroup]:
         if not bastion.enabled:
             return None
+
+        root_device_name = "/dev/xvda"  # This only works for AL2
```
ddl-denis-parnovskiy, line 177: may be move into if/else selector?

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -170,12 +173,16 @@ def provision_vpc(self, stack_name: str, vpc: config.VPC, monitoring_bucket: Opt
...
         if bastion.ami_id:
             region = cdk.Stack.of(self.scope).region
             machine_image = ec2.MachineImage.generic_linux(
                 {region: bastion.ami_id},
-                user_data=ec2.UserData.custom(bastion.user_data),
+                user_data=ec2.UserData.custom(bastion.user_data) if bastion.user_data else None,
```
ddl-denis-parnovskiy, line 182: Hmmm. So it did not work with AMI without user_data

Now we can specify ami_id without user_data. 

It is the change we were talking about but it strange to see it only in this place.

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
steved: Heh sure, updated.

cdk/domino_cdk/provisioners/vpc.py
steved: Well, this is the bastion, so there's nothing "running" on it necessarily. I think it's ok for the bastion to have optional user data for a custom AMI.

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
```diff
@@ -8,6 +8,10 @@
 
 from domino_cdk import config
 
+from ..ami import root_device_mapping
+
+NodeGroup = TypeVar("NodeGroup", bound=config.EKS.NodegroupBase)
```
Secretions, line 13: Might be worth keeping some trace of the base name (`NodegroupCfg` or `CNodegroup` or something) just so it's more obvious looking at the code that it's not a CDK object (which was why everything was imported with `config`--it was getting super confusing differentiating things like `EKS`-the-cfg-object from `eks`-the-cdk-module the way everything was originally imported). Also (super minor nit, sorry), I don't think there's anywhere that capitalizes the `G` in the cdk libs.

Secretions (approved):

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
Secretions: Though apparently we do capitalize it in one of the scopes here: https://github.com/dominodatalab/cdk-cf-eks/blob/243b19410e79bb630cb11cc1a2358e88d441d16a/cdk/domino_cdk/provisioners/eks/eks_nodegroup.py#L134

steved: Lmk what you think of 9bd5fc5ceeb728053d2910a3c84ea25ecdb1cc59?

cdk/domino_cdk/provisioners/vpc.py
ddl-denis-parnovskiy: AFAIR, if we specify custom AMI, 

```
        user_data: ... - Custom script for user_data, ran by cloud_init on node startup. When using the
                         default EKS AMI, this does not replace the default user_data. Your custom script
                         will be injected _before_ the default one, and both will be ran. However, when
                         specifying a custom AMI, this will be the *only* user_data script in use.
```

Does it work?

Can we use custom AMI and not to specify user data and still get something running? 

I did such experiment but I do not remember the outcome already.

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
Secretions: WFM

cdk/domino_cdk/provisioners/vpc.py
steved: Sorry, this change is only for the bastion which doesn't need to run anything from user-data to function correctly. For EKS nodegroup nodes you're definitely correct, we should require user-data.

PR #47, opened 2021-07-20
CI Deploy/Release Support + Minor Pod CIDR Fix

Secretions:
* Renames the testing workflow from `python-app.yml` to `tests.yml`
* Automatically creates and adds release artifacts to github releases
* When a PR has the `deploy-test` label, it will provision via `cdk deploy`, install domino, the uninstall domino and destroy the cluster
* When PR is in ready for review (non-draft), it CI will _fail_ if it hasn't been ran with the deploy option
  * After this merges, we'll require this test to finish successfully
  * In non-draft mode, you can still iterate w/o doing deploys and have CI succeed
  * `deploy-test` does still work in draft mode you if you want to use it
* Explicit template generator variables for all the things required for _every_ install
  * Partially to support this automation, partly for when working manually with it
* Option to disable monitoring bucket (avoiding destroy bug being fixed elsewhere)
* Uploads _entire_ working directory _including repository contents_ as an artifact, stored for 14 days.
  * We create some extra diagnostic data to help research failures
  * We could narrow it down to just the diagnostic data, but (particularly since artifacts are zipped) the storage difference is negligible. Means we can sanity check things we don't necessarily expect to change, added diagnostic data is already picked up, things we didn't realize we need will be there, etc.

A few things in support of pod cidrs tacked on:

* Normalize subnet names (pod cidr subnets had different naming scheme than private/public)
* Calculate POD cidr subnets from supernet
* Add proper security group for pod cidr subnet


cdk/domino_cdk/config/install.py
```diff
@@ -0,0 +1,62 @@
...
+    registry_password: some-password - Password for Domino quay.io image repoistories
+    overrides: <dict/hash> - Overrides of Domino Installer (fleetcommand-agent) configuration.
+    """
+
+    access_list: List[str]  # TODO: What should this variable be? cidr_access_list? loadbalancer_source_ranges?
+    acm_cert_arn: str
+    hostname: str
+    gcr_credentials: str
+    registry_username: str
+    registry_password: str
```
Secretions, line 29: The motivation for these variables is largely to support the automation, and visibility of things that need to be done for every deploy when operating CDK without the deployer.

cdk/util.py
```diff
@@ -44,6 +46,16 @@ def parse_args():
...
     )
+    template_parser.add_argument("--registry-username", help="Quay.io Registry Username", default=None)
+    template_parser.add_argument("--registry-password", help="Quay.io Registry Password", default=None)
+    template_parser.add_argument("--gcr-json-creds", help="GCR Json Credentials", default=None)
+    template_parser.add_argument("--acm-cert-arn", help="ACM Cert ARN", default="__FILL__")
+    template_parser.add_argument(
+        "--hostname", help="Hostname for deployement (ie domino.example.com)", default="__FILL__"
+    )
+    template_parser.add_argument(
+        "--disable-flow-logs", help="Disable monitoring bucket (temporary option)", action="store_true", default=False
```
Secretions, line 57: I've encountered some unreliability when destroying the monitoring bucket. It's being fixed elsewhere, but disabling it here for now.

steved (approved):

Secretions:
@steved Feel free to review for real. :)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,130 @@
...
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Setup nodejs
+      uses: actions/setup-node@v2
+      with:
+        node-version: 16
+    - name: Install aws-cdk
+      run: npm install -g aws-cdk@1.110.1
```
steved, line 30: Probably a follow-up, but I wonder if we can simplify the version constraint sharing with setup.py, etc.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,130 @@
...
+    - name: Test with pytest
+      run: |
+        coverage run -m pytest tests
+
+    - name: Coverage report
+      run: |
+        coverage report
+    - name: Create/lint default config
+      env:
+        AWS_ACCOUNT_ID: ${{ secrets.INFRA_ACCOUNT_ID }}
```
steved, line 53: Let's use one of the accounts that are wiped nightly? Can help you set it up before merging

.github/workflows/tests.yml
```diff
@@ -0,0 +1,130 @@
...
+        GITHUB_SHA: ${{ github.sha }}
+        REGISTRY_USERNAME: ${{ secrets.REGISTRY_USERNAME }}
+        REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
+        GCR_JSON_CREDS: ${{ secrets.GCR_JSON_CREDS }}
+        ACM_CERT_ARN: ${{ secrets.ACM_CERT_ARN }}
+        BASE_DOMAIN: ${{ secrets.BASE_DOMAIN }}
+      run: |
+        export NAME=cdk-deploy-$(echo $GITHUB_SHA | cut -c1-6)
```
steved, line 61: since I had to look it up for `echo TAG=${GITHUB_REF:10}` line, might as well use `${GITHUB_SHA:0:5}`?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,130 @@
...
+        cdk deploy --require-approval never --outputs-file outputs.json
+    - name: Prepare Domino Install
+      if: contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master'
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.INFRA_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.INFRA_ACCESS_KEY }}
+      run: |
+        $(jq -r ".[].ekskubeconfigcmd" outputs.json) --kubeconfig ./kubeconfig
+        jq -r ".[].agentconfig" outputs.json > agent_template.yaml
+        docker run --rm -v $(pwd):/cdk quay.io/domino/deployer:v42 python -m fleetcommand_agent init --full -t /cdk/agent_template.yaml -f /cdk/domino.yml
```
steved, line 103: Why deployer vs. fc-agent? Should we extract the tag somewhere or document when/if to update it?

cdk/domino_cdk/agent.py
```diff
@@ -124,4 +127,20 @@ def generate_install_config(
...
         )
 
+    if install.registry_username:
+        agent_cfg["private_docker_registry"] = {
+            "server": "quay.io",
+            "username": install.registry_username,
+            "password": install.registry_password,
+        }
+
+    if install.gcr_credentials:
```
steved, line 137: I've seen a few people have started recommending the embedded charts over GCR, does that simplify this a bit?

cdk/util.py
```diff
@@ -45,6 +47,16 @@ def parse_args():
         help="KMS Key arn to encrypt kubernetes secrets, generated if not provided",
         default=None,
     )
+    template_parser.add_argument("--registry-username", help="Quay.io Registry Username", default=None)
+    template_parser.add_argument("--registry-password", help="Quay.io Registry Password", default=None)
+    template_parser.add_argument("--gcr-json-creds", help="GCR Json Credentials", default=None)
+    template_parser.add_argument("--acm-cert-arn", help="ACM Cert ARN", default="__FILL__")
+    template_parser.add_argument(
+        "--hostname", help="Hostname for deployement (ie domino.example.com)", default="__FILL__"
```
steved, line 55:
```suggestion
        "--hostname", help="Hostname for deployment (ie domino.example.com)", default="__FILL__"
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,130 @@
...
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.INFRA_ACCESS_KEY }}
+      run: |
+        docker run --rm -v $(pwd):/cdk -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY -e KUBECONFIG=/cdk/kubeconfig quay.io/domino/deployer:v42 python -m fleetcommand_agent run -f /cdk/domino.yml
+    - name: Uninstall Domino
+      if: always() && (contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master')
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.INFRA_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.INFRA_ACCESS_KEY }}
+      run: |
+        docker run --rm -v $(pwd):/cdk -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY -e KUBECONFIG=/cdk/kubeconfig quay.io/domino/deployer:v42 python -m fleetcommand_agent destroy -f /cdk/domino.yml
```
steved, line 117: Similar to our internal stuff, can we store cluster state / logs somewhere for posterity?

.github/workflows/tests.yml
Secretions: Sure. Will static IAM perms get wiped? If not, just lemme know which account is best and I can probably just set it up myself.

.github/workflows/tests.yml
Secretions: The agent doesn't have awscli in it, so the agent image can't be used to do the install phase since the kubeconfig won't be able to auth. Since the agent isn't ever used to install eks from an external source in practice (without the deployer) and at least needs some minor modification (refining the kubeconfig-searching code and come up with some new operational practices), it didn't seem worth adding awscli to the image right this second when there was an alternative available.

Also had some notions of adding a force destroy afterward later, and that may be in the deployer image already--though with moving it into an account that gets wiped regularly I don't really care so much about that anymore.

cdk/domino_cdk/agent.py
Secretions: Yeah, that's a good idea. Takes the required creds down by one.

.github/workflows/tests.yml
Secretions: I can punt them to an s3 bucket or something, but I think I'd rather make a follow-up ticket for that with some thought to cleanup and visibility. There's certainly less of an immediate need for it, without appdevs/cs/etc. also using it for test deploys and it only ever testing/validating work infra devs are already actively working on.

Will likely mostly be useful when there are intermittent deployment errors, particularly with the agent being static and the common case of failures (probably) being poking around in the CloudFormation UI.

steved: Ah I was hoping it'd be something built-in-ish like https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts, but I'm fine with a follow-up and you're right that there's less need.

.github/workflows/tests.yml
Secretions: It's setup in Delta now (and a day has passed and it still works :) ).

cdk/domino_cdk/agent.py
Secretions: I've nuked the gcr creds and am now just statically setting `{"version": 3, cache_path: "charts"}` . I wasn't more specific on the dir since the root is different between the deployer and the agent, but should be relative to the current working dir.

.github/workflows/tests.yml
Secretions: I went ahead and implemented this. I initially was uploading specific files that are useful, but the remainder takes up so little space I decided to just upload everything on the off chance it ends up being useful.

steved (approved):

PR #49, opened 2021-07-26
PLAT-2502: Configurable Subnet Sizes

Secretions:
The `cidr_mask` variable for the public/private subnets has been hardcoded. This makes it configurable.


cdk/domino_cdk/config/template.py
```diff
@@ -78,6 +78,8 @@ def add_nodegroups(name, count, min_size, instance_types, labels, disk_size=100,
         id=None,
         create=True,
         cidr='10.0.0.0/16',
+        public_cidr_mask=27,
```
ddl-denis-parnovskiy, line 81: In 208.128.0.0/**11** and **255.224.0.0** "11" is called prefix and "255.224.0.0" is called mask.

cdk/domino_cdk/config/vpc.py
```diff
@@ -11,6 +11,10 @@ class VPC:
     id: vpc-abc123 - VPC id when using an existing VPC
     cidr: 10.0.0.0/16 - Primary CIDR range for VPC
                         NOTE: EKS needs _lots_ of IPs
+    private_cidr_mask: 19 - Size of private subnets. Should be large enough
```
ddl-denis-parnovskiy, line 14: I'm confused. Is 19 a prefix, or a mask or a network size? It looks like a prefix but is called mask on left side and size of subnet on the right size. The network size is reverse logarithmic proportional to the prefix length. I.e. prefix length 32 means network size of 1.

cdk/domino_cdk/config/template.py
Secretions: I believe `208.120.0.0` is the prefix, `255.224.0.0` us the subnet mask, and `11` is just the cidr notation version of 255.224.0.0 (ie you could read it as `208.128.0.0/255.224.0.0`--they mean the same thing).

Anyhow, the choice is naming is to reflect the api call to `SubnetConfiguration`, which uses `cidr_mask` to name its value (see vpc provisioner in this same PR). I don't believe the terminology here is inaccurate (subnet mask given as cidr value), and would prefer to keep the name mapping the same as the API.

cdk/domino_cdk/config/vpc.py
Secretions: Subnet size in CIDR notation. I updated the phrasing to specify that it's CIDR size.

steved (approved):

PR #50, opened 2021-07-27
support spot instances for unmanaged node pools and add istio-compatibility

steved:
Allow usage of spot instances for unmanaged pools. The priority/capacity is done by providing an ordered list to instance_types.

https://dominodatalab.atlassian.net/browse/PLAT-2050

Istio compatibility is added to ensure ELB provisioning is done correctly.


Secretions (approved):

cdk/domino_cdk/config/template.py
```diff
@@ -28,6 +29,9 @@ def config_template(
         disk_size = 100
         platform_instance_type = "m5.4xlarge"
 
+    if istio_compatible:
```
Secretions, line 32: Dunno if this is less clear / emphasizes the wrong things, but could save a line and make this `if dev_defaults or istio_compatible:`

Secretions (approved):

cdk/domino_cdk/config/template.py
steved: Going to leave it as-is for clarity for now

ddl-denis-parnovskiy: I really do not like this
```
do
if stuff, undo and do differently
```
style

For me it is 3 different configs and 2 bits to chose one.

So I would write
```
normal_config = {
    max_nodegroup_azs = 3
    destroy_on_destroy = False
    disk_size = 1000
    platform_instance_type = "m5.2xlarge"
}

dev_config = {
        max_nodegroup_azs = 1
        destroy_on_destroy = True
        disk_size = 100
        platform_instance_type = "m5.4xlarge"
  }

normal_config_with_istio = {
    max_nodegroup_azs = 3
    destroy_on_destroy = False
    disk_size = 1000
    platform_instance_type = "m5.4xlarge"
}

if not dev and not istio:
  config = normal_config
elif not dev and istio:
  config = normal_config_with_istio
else:
  config = dev_config
```

code to mutate something already done in the previous lines to save a line is not so readable. If you do this in scala, you would probably use val to make stuff immutable and match to select one of the values.

ddl-denis-parnovskiy: It is like "common law". You have original law and you have a myriad of amendments. In the end you need a lawyer to get through.

ddl-denis-parnovskiy: Note that here it is not a "builder" pattern, where you start with something and then add stuff.

cdk/domino_cdk/config/template.py
```diff
@@ -138,19 +142,46 @@ def add_nodegroups(name, count, min_size, instance_types, labels, disk_size=100,
...
+                            "annotations": {"service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "ssl"},
+                        },
+                    }
+                }
+            }
+        }
+
+    if dev_defaults:
+        if not install.get("services"):
+            install["services"] = {}
```
ddl-denis-parnovskiy, line 175:
```
install["services"] = {
  "nucleus": {
  ...
  }
}
```

?

cdk/domino_cdk/config/template.py
```diff
@@ -138,19 +142,46 @@ def add_nodegroups(name, count, min_size, instance_types, labels, disk_size=100,
...
+                            "targetPorts": {"http": "http", "https": "https"},
+                            "annotations": {"service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "ssl"},
+                        },
+                    }
+                }
+            }
+        }
+
+    if dev_defaults:
+        if not install.get("services"):
```
ddl-denis-parnovskiy, line 174: if "services" not in install:

?

ddl-denis-parnovskiy: You also can do
```
nimbus_services = {
  "services": {
    "nimbus": {
    ...
  }
}

install = DominoCdkUtil.deep_merge(install, nimbus_services)

steved: I like the deep_merge one, updated

cdk/domino_cdk/config/template.py
steved: Updated :+1:

PR #54, opened 2021-08-03
add FUSE resource labels to ASGs by default

steved:
Also propagate taints to the ASGs for unmanaged node groups.


Secretions (approved):
lgtm!

PR #55, opened 2021-08-03
ensure nginx-ingress does not use proxy-protocol with istio

steved:
I missed the service annotation for the proxy protocol leading to a mismatch between the nginx config and ELB.

### DIFF
--- a/cdk/domino_cdk/agent.py
+++ b/cdk/domino_cdk/agent.py
@@ -120,14 +120,15 @@ def generate_install_config(
     }
 
     if install.istio_compatible:
+        agent_cfg["istio"] = {
+            "enabled": True,
+            "install": True,
+            "cni": False,
+        }
+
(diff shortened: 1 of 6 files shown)


cdk/domino_cdk/agent.py
```diff
@@ -111,16 +111,70 @@ def generate_install_config(
...
+            {
+                "istio": {
+                    "enabled": True,
+                    "install": True,
+                    "cni": False,
+                },
+                "services": {
+                    "nginx_ingress": {
+                        "chart_values": {
+                            "controller": {
```
Secretions, line 134: Wonder if it would be cleaner if you just had `agent_cfg["istio"] = {<istio config>}`, and then did the deep merge on just the contents of controller here?

Secretions (approved):

PR #56, opened 2021-08-07
PLAT-2499: Make bucket roles strict, names (sanely) configurable

Secretions:
* Make it so that bucket _types_ are statically configured
* Adds an option to modify the bucket name in case we need to
* Adds a feature to make certain config values hidden
* Makes the bucket name override hidden so we don't clutter the config
* Removes the `-rcX` suffix from the version in code, since CI takes care of that for us now

### DIFF
--- a/cdk/domino_cdk/config/s3.py
+++ b/cdk/domino_cdk/config/s3.py
@@ -69,8 +69,6 @@ class S3:
     def from_0_0_0(c: dict):
         return from_loader(
             "config.s3",
-            S3(
-                buckets=S3.BucketList.load(c.pop("buckets"))
-            ),
+            S3(buckets=S3.BucketList.load(c.pop("buckets"))),
             c,
         )
--- a/cdk/domino_cdk/config/s3.py
+++ b/cdk/domino_cdk/config/s3.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 from typing import Dict, Optional
 
-from domino_cdk.config.util import from_loader
+from domino_cdk.config.util import check_leavins, from_loader
 
 
 @dataclass
@@ -9,9 +9,12 @@ class S3:
     """
     Map of buckets to create and provide IAM access from the EKS cluster.
 
-    The monitoring bucket is optionally created to store various operational logs: VPC flow logs, S3 and ELB access logs.
+    "blobs", "logs", "backups" and "registry" buckets are MANDATORY.
+
+    The "monitoring" bucket is optionally created to store various operational logs: VPC flow logs, S3 and ELB access logs.
 
     Bucket parameters:
+    name: "my-bucket-name" - Optional name to override default bucketname of deployname-bucket-type
     auto_delete_objects: true/false - Delete entire contents of bucket when destroying the CloudFormation stack
     removal_policy_destroy: true/false - Delete bucket when destroying stack. If auto_delete_objects is false,
                                          destroys will fail unless buckets are emptied.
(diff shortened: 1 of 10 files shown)


cdk/tests/unit/config/test_s3.py
```diff
@@ -21,17 +31,32 @@
...
+        monitoring=S3.BucketList.Bucket(
+            auto_delete_objects=True, removal_policy_destroy=True, sse_kms_key_id=sse_kms_key_id
+        ),
+    )
 )
 
 
 class TestConfigS3(unittest.TestCase):
     def test_from_0_0_0(self):
+        return True
```
steved, line 54: ✂️ ?

cdk/domino_cdk/config/s3.py
```diff
@@ -1,48 +1,86 @@
...
     """
     Map of buckets to create and provide IAM access from the EKS cluster.
 
-    The monitoring bucket is optionally created to store various operational logs: VPC flow logs, S3 and ELB access logs.
+    "blobs", "logs", "backups" and "registry" buckets are MANDATORY.
+
+    The "monitoring" bucket is optionally created to store various operational logs: VPC flow logs, S3 and ELB access logs.
 
     Bucket parameters:
+    name: "my-bucket-name" - Optional name to override default bucketname of deployname-bucket-type
```
steved, line 18: Is this something we really want to support configuring or just an easy toss-in?

steved (commented):
Can we do something less implicit on the [agent side](https://github.com/dominodatalab/cdk-cf-eks/blob/33471a776f213c6215d22f113808db047bf9f68d/cdk/domino_cdk/agent.py#L36-L60) as well?

cdk/domino_cdk/config/s3.py
ddl-denis-parnovskiy: We can configure buckets in the deployment config. So here is a must.

I'm more concerned about 2 sources of truth: deployment config and provisioning config (cdk config). There is stuff to be configured in both like bucket names, k8s namespaces.

It would be good to have one config instead of multiple.

ddl-denis-parnovskiy (approved):

ddl-denis-parnovskiy:
Perhaps I'm blind but I do not see what it is about "bucket roles strict"

cdk/domino_cdk/config/s3.py
Secretions: Since s3 buckets are globally namespaced (and occasionally s3 is slow to release names when iterating), I thought it necessary to have a safety valve with the bucket names. I _don't_ want people configuring it, which is why I added the concept of the hidden option for it.

Perhaps documenting it there is still too visible? I kind of fear that _we'll_ forget we can even do it in the head of the moment, though. :)

Secretions:
> Perhaps I'm blind but I do not see what it is about "bucket roles strict"

The bucket keys were previously both the name of the bucket and its name, conflated, and were a completely arbitrary dict that could have infinite buckets in it. So long as the places where we hardcoded the keys we wanted were happy, you could do whatever. You also couldn't choose bucket names that differ from the default because of this.

Now we have explicit buckets hardcoded to their bucket roles, and the config object is strict in that it's explicitly loading them by the name of their role instead of loosely loading whatever you fed it and letting it fail elsewhere if we're missing something. You cannot load a config that is missing required bucket roles, nor can you load a config with extraneous buckets. If you have a config in memory, it *will* definitely have a config bucket object defined for each required bucket role we have (blobs, logs, etc.).

Secretions:
* Make it so that bucket types are statically configured
* Adds an option to modify the bucket name in case we need to
* Adds a feature to make certain config values hidden
* Makes the bucket name override hidden so we don't clutter the config
* Removes the -rcX suffix from the version in code, since CI takes care of that for us now

PR #58, opened 2021-08-12
do not re-map custom AMIs volumes

steved:
There's always the chance we don't have the right settings (KMS keys,
encryption, etc.) so for custom AMIs we just take the default volume
mappings.


cdk/domino_cdk/config/iam.py
```diff
@@ -347,8 +347,6 @@ def do_cf():
             "eks:DescribeNodegroup",
             "eks:TagResource",
             "eks:UntagResource",
-            "ec2:RevokeSecurityGroupEgress",
-            "ec2:RevokeSecurityGroupIngress",
```
steved, line 351: This is a duplicate of these lines:
https://github.com/dominodatalab/cdk-cf-eks/blob/4d06375c75a5887c85bcdce51c95993262266e32/cdk/domino_cdk/config/iam.py#L396-L397

https://github.com/dominodatalab/cdk-cf-eks/blob/4d06375c75a5887c85bcdce51c95993262266e32/cdk/domino_cdk/config/iam.py#L413-L414

ddl-denis-parnovskiy: Should we go opposite and remove * permissions?

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
```diff
@@ -277,16 +275,15 @@ def _handle_user_data(
...
-        opts = {
+    def _launch_template(self, scope, name: str, ng: config.eks.T_NodegroupBase, **kwargs) -> ec2.LaunchTemplate:
+        opts: Dict[Any, Any] = {
             "key_name": ng.key_name,
             "launch_template_name": f"{self.stack_name}-{name}",
-            "block_devices": [
+        }
+
+        if not ng.ami_id:
+            root_device_name = "/dev/xvda"  # This only works for AL2
```
ddl-denis-parnovskiy, line 285: How they do remap if they use AL2 cloned image?

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -193,20 +189,30 @@ def provision_bastion(self, name: str, bastion: config.VPC.Bastion) -> Optional[
...
+            bastion_opts["machine_image"] = ec2.MachineImage.generic_linux(
                 {region: bastion.ami_id},
                 user_data=ec2.UserData.custom(bastion.user_data) if bastion.user_data else None,
             )
-
-            root_device_name = root_device_mapping(self.scope, bastion.ami_id).name
         else:
-            machine_image = ec2.GenericSSMParameterImage(
+            bastion_opts["machine_image"] = ec2.GenericSSMParameterImage(
                 "/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2", ec2.OperatingSystemType.LINUX
```
ddl-denis-parnovskiy, line 202: Why we are using a specific image here instead of default?

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
steved: Something like [packer](https://www.packer.io/docs/builders/amazon/ebs) can do this really easily.

cdk/domino_cdk/provisioners/vpc.py
steved: I don't think there is a default? This is the latest Amazon Linux 2 AMI.

cdk/domino_cdk/config/iam.py
steved: @Secretions preference?

Secretions: The only two things behind the wildcard are Authorize and Revoke, so the wildcard is probably fine.

Secretions (approved):

cdk/domino_cdk/config/iam.py
ddl-denis-parnovskiy: If there is only one additional operation that we intent to address with *, then definitely list it explicitly.

```
 "ec2:Authorize SecurityGroupEgress", 
 "ec2:Authorize SecurityGroupIngress", 
 "ec2:RevokeSecurityGroupEgress", 
 "ec2:RevokeSecurityGroupIngress", 
```

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
ddl-denis-parnovskiy: There is a neat AWS image builder as AWS online service. It allows to customize an existing AMI in AWS image repo by adding custom layers, like SSM. It doesn't run packer.

https://docs.aws.amazon.com/imagebuilder/latest/userguide/how-image-builder-works.html

Are we saying that if they build a customized AL2 image and need to more root device name to '/dev/xda' then they have to abandon AWS image building pipeline aaS and use offline 3rd party tool plus maintain their own building pipeline?

cdk/domino_cdk/provisioners/vpc.py
ddl-denis-parnovskiy: I mean if we do not specify an image then latest AL2 is automatically picked. 

Or, IIRC, there is a function to pick default image by family.

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
ddl-denis-parnovskiy: After playing with EC3 image builder, 

on one hand, the disk size is an easy option when creating an image.

on the other hand, they will lose an ability to use the same image in different deployments by scaling the sizes from config.yaml.

on third hand, we do not give them ability to scale CPU/memory anyway.

Interesting, though, that LaunchTamplate API allows CPU selection https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.CfnLaunchTemplate.CpuOptionsProperty.html it doesn't allow memory selection.

Ideal solution, IMHO, is to 
1. Give them an option to scale CPU 
2. Give them a config option to explicitly indicate the device names and sizes. See, AL2 has more disks that /dev/xda https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html

If we cannot do this now, at least document that with custom AMI disk sizes are also not supported.

cdk/domino_cdk/provisioners/vpc.py
steved: This is that function IMO?

cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
steved: Let me know what you think of 4676f075ecef58ab29a7f2c97d62e5f98df73294.

PR #61, opened 2021-08-16
PLAT-37: Install coredns/kube-proxy/vpc-cni addons

Secretions:
* Install coredns, kube-proxy and vpc-cni as "EKS addons"
  * These are all installed by default, but unmanaged
  * On EKS upgrades, these services won't get upgraded unless they're managed as addons
* Uses boto3 to get the most current version of any given addon
  * You can leave the version field blank, but it won't get updated
  * There is no "use the latest" function in cloudformation/cdk
  * Have an alternate version that uses 3 (!!!) custom resources instead of boto3: https://github.com/dominodatalab/cdk-cf-eks/pull/62
    * The CRs just feel too icky to me........
    * But somebody tell me if I should just give in and do them........ ugh.
* Adds new runtime requirement of valid aws creds for aws synth, which was only required for deploy previously

### DIFF
--- a/cdk/app.py
+++ b/cdk/app.py
@@ -1,10 +1,19 @@
 #!/usr/bin/env python3
+import boto3
 from aws_cdk import core
 from ruamel.yaml import SafeLoader
 from ruamel.yaml import load as yaml_load
 
 from domino_cdk.config import config_loader
 from domino_cdk.domino_stack import DominoStack
+
+sts = boto3.client("sts")
+try:
+    sts.get_caller_identity()
+except Exception:
+    print("WARNING: Domino CDK App requires valid AWS credentials!\n")
+    raise
+
 
 app = core.App()
 
--- a/cdk/domino_cdk/aws_configurator.py
+++ b/cdk/domino_cdk/aws_configurator.py
@@ -25,7 +25,6 @@ class DominoAwsConfigurator:
         self.eks_cluster = eks_cluster
 
         self.install_calico()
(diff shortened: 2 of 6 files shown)


ddl-denis-parnovskiy (approved):

cdk/app.py
```diff
@@ -1,11 +1,20 @@
...
 from aws_cdk import core
 from ruamel.yaml import SafeLoader
 from ruamel.yaml import load as yaml_load
 
 from domino_cdk.config import config_loader
 from domino_cdk.domino_stack import DominoStack
 
+sts = boto3.client("sts")
+try:
+    sts.get_caller_identity()
```
steved, line 12: I have no idea if this is true, but in theory could we have credentials that work without STS access?

steved (approved):

cdk/app.py
Secretions: https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

```
No permissions are required to perform this operation. If an administrator adds a policy to your IAM user or role that explicitly denies access to the sts:GetCallerIdentity action, you can still perform this operation. Permissions are not required because the same information is returned when an IAM user or role is denied access. To view an example response, see I Am Not Authorized to Perform: iam:DeleteVirtualMFADevice in the IAM User Guide.
```

Secretions: Doesn't seem like this is a concern.

PR #64, opened 2021-08-18
PLAT-2407: Switch GPU to g4dn.xlarge

fraenkel:
Switch the GPU nodes to g4dn.xlarge


cdk/domino_cdk/config/template.py
```diff
@@ -80,7 +80,7 @@ def add_nodegroups(name, count, min_size, instance_types, labels, disk_size=100,
         "gpu",
         gpu_nodegroups,
         0,
-        ["p3.2xlarge"],
+        ["g4dn.xlarge"],
```
steved, line 83: unless we're comfortable addressing PLAT-2441, we should do
```
["g4dn.xlarge" if dev_defaults else "p3.2xlarge"]
```

fraenkel: or we use larger g4dn instances which are also cheaper than p3.

steved (approved):

PR #71, opened 2021-09-07
Switch to GP3 (PLAT-1118)

ddl-denis-parnovskiy:
See https://dominodatalab.atlassian.net/browse/PLAT-1118 for details

Just changing GP2->GP3 values.

### DIFF
--- a/cdk/domino_cdk/provisioners/vpc.py
+++ b/cdk/domino_cdk/provisioners/vpc.py
@@ -209,7 +209,7 @@ class DominoVpcProvisioner:
                         40,  # note: this requires the AMI root device already be <=40GB
                         delete_on_termination=True,
                         encrypted=True,
-                        volume_type=ec2.EbsDeviceVolumeType.GP2,
+                        volume_type=ec2.EbsDeviceVolumeType.GP3,
                     ),
                 )
             ]
--- a/cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_nodegroup.py
@@ -290,7 +290,7 @@ class DominoEksNodegroupProvisioner:
                         ng.disk_size,
                         delete_on_termination=True,
                         encrypted=True,
-                        volume_type=ec2.EbsDeviceVolumeType.GP2,
+                        volume_type=ec2.EbsDeviceVolumeType.GP3,
                     ),
                 )
             ]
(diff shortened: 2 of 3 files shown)


fraenkel:
need to deploy it

ddl-denis-parnovskiy:
Deployed with cdk deploy

You can see gp3 volumes created, e.g. https://us-west-2.console.aws.amazon.com/ec2/v2/home?region=us-west-2#Volumes:volumeId=vol-0d86134daa8550817;sort=desc:createTime

cdk/domino_cdk/provisioners/vpc.py
```diff
@@ -209,7 +209,7 @@ def provision_bastion(self, name: str, bastion: config.VPC.Bastion) -> Optional[
                         40,  # note: this requires the AMI root device already be <=40GB
                         delete_on_termination=True,
                         encrypted=True,
-                        volume_type=ec2.EbsDeviceVolumeType.GP2,
+                        volume_type=ec2.EbsDeviceVolumeType.GP3,
```
Secretions, line 212: Curious what the upgrade experience is like for the bastion.

ddl-denis-parnovskiy: We do not do re-provisioning, correct?
I.e. bastion is now provisioned with gp3 block storage and this is it.
What upgrade experience differences you expect?

Secretions: We don't reprovision as a matter of course on upgrades, but it's not true that we'll never ever reprovision.

The ideal outcome is that since the bastion is already provisioned, cloudformation doesn't bother reprovisioning the bastion despite the change since it already exists (and we don't care much about the bastion and its disk type). I don't expect this to happen, and wouldn't want to waste any time engineering it to happen.

What I assume will happen is that CloudFormation will destroy the bastion and create a new one. This is fine, as people really shouldn't store important stuff on the bastion and there will be probably minimal people that go through this upgrade over time. But it'd be nice to confirm this happens as expect, that we don't run into unexpected errors and can still connect to the deployment afterward (ie output w/the bastion ip address isn't cached on the deployer side or something).

ddl-denis-parnovskiy:
OK.let me run an experiment like this

cdk/domino_cdk/provisioners/vpc.py
ddl-denis-parnovskiy: OK.let me run an experiment like this

ddl-denis-parnovskiy: Switching the code from GP3 back to GP2 did not trigger reprovisioning of Bastion.

I'll try GP2-GP3 path

ddl-denis-parnovskiy: Yes, CF scraps the old bastion instance and deploys a new one

fraenkel (approved):

PR #73, opened 2021-09-20
PLAT-1118: Add ebs-csi policy for Nodegroups

fraenkel:


cdk/domino_cdk/provisioners/eks/eks_iam.py
```diff
@@ -83,10 +84,170 @@ def provision(self, stack_name: str, cluster_name: str, r53_zone_ids: List[str],
...
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteVolume"
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/kubernetes.io/cluster/*": "owned"
```
Secretions, line 205: Seems like we might be able to restrict it to just our cluster?

cdk/domino_cdk/provisioners/eks/eks_iam.py
```diff
@@ -83,10 +84,170 @@ def provision(self, stack_name: str, cluster_name: str, r53_zone_ids: List[str],
...
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true"
+        }
+      }
+    },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ec2:DeleteVolume"
```
ddl-denis-parnovskiy, line 188: Use the force. https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_multi-value-conditions.html

You can combine string like.

cdk/domino_cdk/provisioners/eks/eks_iam.py
```diff
@@ -83,10 +84,170 @@ def provision(self, stack_name: str, cluster_name: str, r53_zone_ids: List[str],
...
+      ],
+      "Resource": "*",
+      "Condition": {
+        "StringLike": {
+          "ec2:ResourceTag/ebs.csi.aws.com/cluster": "true"
```
ddl-denis-parnovskiy, line 181: Please limit it by domino-deploy tag. A code error will lead to major data loss.

cdk/domino_cdk/provisioners/eks/eks_iam.py
fraenkel: done

cdk/domino_cdk/provisioners/eks/eks_iam.py
fraenkel: it gets confusing since its not quite the same for snapshot... its easier to read to group like actions.

cdk/domino_cdk/provisioners/eks/eks_iam.py
fraenkel: not sure that comes out the other end.  I will have to see what we can provide the csi provisioner

cdk/domino_cdk/provisioners/eks/eks_iam.py
```diff
@@ -83,10 +83,63 @@ def provision(self, stack_name: str, cluster_name: str, r53_zone_ids: List[str],
...
+                        "ec2:DetachVolume",
+                        "ec2:ModifyVolume",
+                        "ec2:DescribeAvailabilityZones",
+                        "ec2:DescribeInstances",
+                        "ec2:DescribeSnapshots",
+                        "ec2:DescribeTags",
+                        "ec2:DescribeVolumes",
+                        "ec2:DescribeVolumesModifications",
+                    ],
+                    resources=["*"],
```
ddl-denis-parnovskiy, line 103: Any chance to limit it, say by 

conditions={"StringNotEqualsIfExists": {"ecr:ResourceTag/domino-deploy-id": stack_name}}

fraenkel: Trying with the correct condtion.
Look at what EKS Worker Nodes get.... the above is a bit pointless since it isn't being verified in some cases.

fraenkel: This interferes with ASGs if you pick the wrong tag.

fraenkel: verified a CDK deploy does work

fraenkel:
Install domino and destroy CDK are failing. However from the other direction (platform-apps), it all works fine.

Secretions:
It's a mismatch between the pinned deployer and the develop.latest tag on nucleus. I think we should either pin the deployer to v46 or platform-apps's develop.latest

fraenkel:
I don't believe that will work. Its going to cause the gp2 (in-tree) operator which is gonna fail due to these new policies since you have me matching on very specific tags.

Secretions:
Tested it here and it seems to work: https://github.com/dominodatalab/cdk-cf-eks/pull/75

Secretions (approved):

PR #76, opened 2021-09-30
PLAT-2802: Set metrics server up to be installed in agent cfg

Secretions:


.github/workflows/tests.yml
```diff
@@ -60,7 +68,7 @@ jobs:
         BASE_DOMAIN: ${{ secrets.DELTA_BASE_DOMAIN }}
       run: |
         export NAME=cdk-deploy-${GITHUB_SHA:0:6}
-        ./util.py generate_config_template --name $NAME --aws-region=us-west-2 --aws-account-id=$AWS_ACCOUNT_ID --dev --registry-username $REGISTRY_USERNAME --registry-password $REGISTRY_PASSWORD --hostname $NAME.$BASE_DOMAIN --acm-cert-arn $ACM_CERT_ARN --disable-flow-logs > config.yaml
+        ./util.py generate_config_template --name $NAME --aws-region=us-west-2 --aws-account-id=$AWS_ACCOUNT_ID --dev --platform-nodegroups 2 --registry-username $REGISTRY_USERNAME --registry-password $REGISTRY_PASSWORD --hostname $NAME.$BASE_DOMAIN --acm-cert-arn $ACM_CERT_ARN --disable-flow-logs > config.yaml
```
steved, line 71: Why 2 nodegroups?

.github/workflows/tests.yml
```diff
@@ -28,6 +28,14 @@ jobs:
       uses: actions/setup-node@v2
       with:
         node-version: 16
+    - name: Determine deployer image
+      env:
+        PR_BODY: ${{ github.event.pull_request.body }}
+      run: |
+        export PR_DEPLOYER_IMAGE=$(echo $PR_BODY | grep -oP "deployer_image: \K\S+")
```
steved, line 35: cute, can we document this and the deploy-test label in the readme?

steved (approved):

.github/workflows/tests.yml
Secretions: Sure. I'll get it in followup just because I'm going to be redoing the readme anyway.

.github/workflows/tests.yml
Secretions: Just simplest way to get us two default nodes without either doing some template override/sed shenanigans or adding more args to the template generator, and I'm only doing this as a temporary thing so I didn't want it to be the need that drives a `--platform-min` type arg.

ddl-denis-parnovskiy (approved):

PR #78, opened 2021-12-02
Add eks:DescribeUpdate perms

Secretions:
Some recent upgrade testing has shown a few missing perms.


Secretions:
Delete failure was spurious network interface sticking around, not likely related to perms.

steved (approved):

PR #83, opened 2021-12-22
PLAT-3779: Add DescribeInstanceStatus IAM policy

senthilddl:
### Link to JIRA

[PLAT-3779](https://dominodatalab.atlassian.net/browse/PLAT-3779)

### What issue does this pull request solve?

Adds `DescribeInstanceStatus` IAM policy to EC2 instances which is needed for [Shoreline](https://shoreline.io/)

### DIFF
--- a/cdk/domino_cdk/config/iam.py
+++ b/cdk/domino_cdk/config/iam.py
@@ -338,6 +338,7 @@ def generate_iam(stack_name: str, aws_account_id: str, region: str, manual: bool
             "ec2:DescribeAvailabilityZones",
             "ec2:DescribeImages",
             "ec2:DescribeInstances",
+            "ec2:DescribeInstanceStatus",
             "ec2:DescribeLaunchTemplates",
             "ec2:DescribeSecurityGroups",
             "ec2:DisassociateAddress",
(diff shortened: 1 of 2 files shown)


cdk/domino_cdk/config/iam.py
```diff
@@ -338,6 +338,7 @@ def do_cf():
             "ec2:DescribeAvailabilityZones",
             "ec2:DescribeImages",
             "ec2:DescribeInstances",
+            "ec2:DescribeInstanceStatus",
```
Secretions, line 341: This is actually the deployment policy, not the policy running on the nodes, so I don't think it's actually needed here?

senthilddl: Ah ok, Reverting here.

Secretions: This was actually the opposite reversion that happened (the one this comment is one should go away, the other one should come back. Sorry. :(

senthilddl: My bad, sorry, I read it correctly earlier, not sure what I was thinking.

Secretions (approved):

PR #90, opened 2022-04-05
set kubernetes 1.22 as the default; update cdk libraries to 1.151

steved:
If we want to continue using the official `V1_XX` accessors, we'll need to wait for the latest CDK version. The deploy test will continue to fail due to missing v1.22 support in Domino.

If necessary, I can split this out into two changes.


Secretions:
> If we want to continue using the official `V1_XX` accessors, we'll need to wait for the latest CDK version. The deploy test will continue to fail due to missing v1.22 support in Domino.
> 
> If necessary, I can split this out into two changes.

I think I would personally prefer to split out the v1.22 change and wait for it to actually work before merging it in.

PR #91, opened 2022-04-21
PLAT-4562: Allow nodegroup selection per AZ

Secretions:
Currently, you can't select which AZs a given nodegroup will use, despite the fact that we have a need to pick and choose (ie gpu instance type not available in xx-west-2b despite being in 2a and 2c or what not). This allows us to do so.


cdk/tests/unit/config/test_config.py
```diff
@@ -57,6 +57,28 @@ def test_invalid_version_syntax(self):
...
+
+        c = config_template().render()
+        c["vpc"]["max_azs"] = 3
+        c["eks"]["unmanaged_nodegroups"]["platform-0"]["availability_zones"] = [
+            "us-west-2a",
+            "us-west-2d",
+            "us-west-2c",
+        ]
+        with self.assertRaisesRegex(
+            ValueError, "Nodegroup platform-0 has availability zone us-west-2d exceeding vpc.max_azs count: 3"
```
fraenkel, line 78: the test says 3 AZs, the list has 3, so why does it fail?

cdk/domino_cdk/config/base.py
```diff
@@ -127,6 +128,26 @@ def val(path: str, obj):
...
+        for ng, cfg in self.eks.managed_nodegroups.items():
+            if az_pos(cfg.availability_zones) > self.vpc.max_azs:
+                errors.append(
+                    f"Nodegroup {ng} has availability zone {last_az(cfg.availability_zones)} exceeding vpc.max_azs count of {self.vpc.max_azs}"
+                )
+        for ng, cfg in self.eks.unmanaged_nodegroups.items():
+            if az_pos(cfg.availability_zones) > self.vpc.max_azs:
+                errors.append(
+                    f"Nodegroup {ng} has availability zone {last_az(cfg.availability_zones)} exceeding vpc.max_azs count: {self.vpc.max_azs}"
+                )
```
fraenkel, line 149: this isn't right.
All you care about is that the AZs selected match the AZs we have defined in the VPC.
We can check that the number of AZs selected is less than or equal to the VPC AZs but not really necessary.

fraenkel (changes requested):

cdk/tests/unit/config/test_config.py
Secretions: Other comment explains the reasoning, but it's the presence of `us-west-2d` (the "fourth" AZ) that causes the failure.

cdk/domino_cdk/config/base.py
fraenkel: so an easier way is to just do `len(set(cfg.availability_zones)) > self.vpc.max_azs` no need to do anything fancy.

cdk/tests/unit/config/test_config.py
fraenkel: so I dont see how d is the fourth. There are 3 defined AZs, a, d, c, now if it was a,d,c,X then I agree you have a forth. Asking for a max of 3 doesn't guarantee a,b,c.

cdk/domino_cdk/config/base.py
Secretions: I switched it up so it looks at what AZs are in the configured region, replicates the logic of the vpc, and checks if each nodegroup's AZs are in the projected VPC's subnet AZs. This should cover regions where an AZ letter is skipped.

fraenkel (approved):
sorry about that but I hit this issue before...

PR #100, opened 2022-05-27
HBD-186: Introduce configuration fields to disable resources which aren't needed for data plane

ddl-kevin:
Introduce optional `enabled` field (default `true`) to configurations for following resources:

* S3 buckets
* EFS
* Route53
* Installer configuration

This would allow the deployer to simply set `enabled: false` for each of these resource configurations to deploy a data plane VPC / EKS cluster with identical configuration as our existing dev/prod CDK flavors.

### DIFF
--- a/cdk/domino_cdk/config/efs.py
+++ b/cdk/domino_cdk/config/efs.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from typing import Optional
 
 from domino_cdk.config.util import from_loader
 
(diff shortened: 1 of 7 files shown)


cdk/domino_cdk/config/efs.py
```diff
@@ -28,9 +29,17 @@ class Backup:
...
     backup: Backup
     removal_policy_destroy: bool
+    enabled: bool = True
 
     @staticmethod
-    def from_0_0_0(c: dict):
+    def from_0_0_0(c: dict) -> Optional['EFS']:
+        enabled = c.get("enabled", True)
+        if "enabled" in c:
+            del c["enabled"]
```
Secretions, line 38: The pattern here (for all of the sections) should really be more like:

```
return from_loader(
  "config.thing",
  Thing(
    enable=thing.pop("enable", True),
    existing_variable=thing.pop("existing_variable"),
    [...]
)
```

cdk/domino_cdk/config/install.py
```diff
@@ -27,9 +27,17 @@ class Install:
     registry_password: str
     overrides: dict
     istio_compatible: bool
+    enabled: bool = True
```
Secretions, line 30: For the install section specifically, saying `enabled` feels a bit weird since this section doesn't actually do anything, it's just informational (ie it generates an agent config template as an output).

I wonder if it would make more sense to make this section entirely optional, so you can just load a config without an install block or perhaps one that's explicitly blank or set to null? Or perhaps that should just be the pattern overall—allow certain blocks to be null? Probably prevents a lot of irrelevant stuff from being in the config and validated too.

The config generator util.py uses could have args that prevent those blocks from being generated in the first place, and then in platform-apps the args sent to the config generator in the manifest could be used.

Just a suggestion, but within the pattern we've used I might either rename this variable to something that indicates it's not generating the config, or just omit it and infer from what modules were disabled that we don't need to generate the agent template.

cdk/domino_cdk/config/efs.py
ddl-kevin: The reason I didn't go this route was to prevent having to parse a valid configuration when disabled. If you are ok with using empty sections to indicate disabled, that feels more ergonomic to me and I can make that change. thoughts?

Secretions: Yeah, let's try the empty sections. I think it will make a lot more sense and result in a cleaner (and more relevant) config for the data plane clusters.

Secretions (approved):
I think some of the conditionals could be simplified as mentioned, but otherwise lgtm.

PR #102, opened 2022-06-08
Enable customers to bring their own vpc's

ddl-zacharyblasczyk:
Per documentation https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.aws_ec2/VpcLookupOptions.html this is required to look up a vpc. 

Addresses PLAT-4992


steved:
Closing since this was addressed in #103

PR #103, opened 2022-06-10
Enable customers to bring their own vpc's

Secretions:
Per documentation https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.aws_ec2/VpcLookupOptions.html this is required to look up a vpc. 

Addresses PLAT-4992


Secretions:
Someone else's change that I repackaged, consider this comment an approval as I admin-merge.

PR #105, opened 2022-08-10
remove prune setting for eks cluster

steved:
AWS-CDK's lambda for generic kubernetes manifests doesn't support 1.22 officially so `prune` bails on the missing API versions (like ingress). This occurs during updates to an existing cluster. If/when we update Calico, this may affect how the upgrades from [one version to another works](https://github.com/dominodatalab/cdk-cf-eks/blob/master/cdk/domino_cdk/aws_configurator.py).

See https://github.com/aws/aws-cdk/issues/19843 for more context. https://github.com/aws/aws-cdk/issues/15736#issuecomment-1212926265 does indicate a workaround, but it doesn't seem very palatable.

deployer_image: quay.io/domino/deployer:steved-destroy-deploy.latest

### DIFF
--- a/cdk/domino_cdk/provisioners/eks/eks_cluster.py
+++ b/cdk/domino_cdk/provisioners/eks/eks_cluster.py
@@ -77,6 +77,7 @@ class DominoEksClusterProvisioner:
             default_capacity=0,
             security_group=eks_sg,
             secrets_encryption_key=key,
+            prune=False,  # https://github.com/aws/aws-cdk/issues/19843
         )
 
         # To make sure log cleanup is called after cluster cleanup: cluster depends on custom so custom is guaranteed


steved:
Tested with a successful deploy of 1.22 on master. Adding a new unmanaged node group and `cdk deploy` results in:
```
Received response status [FAILED] from custom resource. Message returned: Error: b'configmap/aws-auth configured\nerror: error retrieving RESTMappings to prune: invalid resource extensions/v1beta1, Kind=Ingress, Namespaced=true: no matches for kind "Ingress" in version "extensions/v1beta1"\n' Logs: /aws/lambda/steved-cdk-1-EksStackNestedStackEk-Handler886CB40B-2ak0Zp34wmv1 at invokeUserFunction (/var/task/framework.js:96:19) at processTicksAndRejections (internal/process/task_queues.js:97:5) at async onEvent (/var/task/framework.js:19:27) at async Runtime.handler (/var/task/cfn-response.js:48:13) (RequestId: 89089f1c-1801-4430-9479-200c70529c8e)
```

With this change, it completes successfully.

Secretions (approved):

PR #106, opened 2022-08-15
PLAT-5302: Update CDK to use 1.1 compatible schema

Secretions:


ddl-dclegg (approved):

cdk/domino_cdk/agent.py
```diff
@@ -69,18 +81,17 @@ def generate_install_config(
             },
         },
         "internal_docker_registry": {
+            "enabled": True,
```
fraenkel, line 84: not a thing

cdk/domino_cdk/agent.py
```diff
@@ -69,18 +81,17 @@ def generate_install_config(
...
         "metrics_server": {"install": True},
         "gpu": {"enabled": True},
-        "helm": {
-            "cache_path": "charts",
-        },
-        "services": {
-            "nginx_ingress": {},
+        "release_overrides": {
+            "nginx-ingress": {},
             "forge": {
```
fraenkel, line 95: forge? shouldn't there be hephaestus?

cdk/domino_cdk/agent.py
Secretions: While you're correct and I've removed it, this inexplicably doesn't cause a validation error.

steved: We're still temporary lenient on this specific value until https://github.com/cerebrotech/platform-apps/pull/5830 is merged:
https://github.com/cerebrotech/platform-apps/blob/develop/src/domino_config/types1_1.py#L641

.github/workflows/tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   test-build-deploy:
     runs-on: ubuntu-latest
     env:
-      DEPLOYER_IMAGE: quay.io/domino/deployer:develop.latest
+      DEPLOYER_IMAGE: quay.io/domino/deployer:plat-5302-cdk.latest
```
fraenkel, line 17: this doesn't look right

PR #109, opened 2022-08-25
Schema 1.1 fix

Secretions:
The change moving private_docker_registry under helm never made it into this.


Secretions:
Tested with an existing install. Turbomerging, apologies.

PR #110, opened 2022-08-31
HBD-500 add ability to create certificates in ACM

ddl-bgreen:


ddl-kevin (approved):
I think this looks good, the test failure may be a flake, but Joaquin probably knows better.

It seems like this lets us configure the cert in either the CP or the DP, depending on the changes we add to platform apps, so that is nice.

PR #111, opened 2022-09-19
Dumb fix for diff false positives on calico

Secretions:
There are some obnoxious utf-8 quotes in the calico manifest for some reason, which get automatically converted to `?` by AWS at some point. Because of this, they _always_ present as a difference when doing `cdk diff`.

This change performs that same conversion, but within our cdk app. Because this is json, I do the same thing they do and convert it to `?` rather than `"`, as otherwise we're put in the position of iterating through every key and value individually to perform the substitutions to do it safely.


steved (approved):
annoying, nice catch!

PR #116, opened 2022-10-21
update lambda to use cfnresponse

steved:
When building a custom response, we were returning the resource property data which could exceed the 4k response limit.


cdk/domino_cdk/provisioners/lambda_files/backup_post_creation_tasks.py
```diff
@@ -40,4 +33,3 @@ def on_delete(event):
         response = client.delete_recovery_point(
```
fraenkel, line 33: we shouldn't set response

fraenkel (approved):
hmm, we dont have tests?

cdk/domino_cdk/provisioners/lambda_files/backup_post_creation_tasks.py
steved: yep, good catch

steved:
Will follow-up here for unit tests: https://github.com/dominodatalab/cdk-cf-eks/pull/119

PR #121, opened 2022-10-31
Add Support for `service_ipv4_cidr` in EKS

philnielsen:
This allows someone to add a custom EKS Service CIDR range. This was opened at the suggestion of our rep from your team, who said you all would consider this enhancement. This configuration option is required if you need network peering with the [default EKS CIDRs](https://docs.aws.amazon.com/eks/latest/APIReference/API_KubernetesNetworkConfigRequest.html#AmazonEKS-Type-KubernetesNetworkConfigRequest-serviceIpv4Cidr) for different services within your account.

Note I do not have the resources or bandwidth to fully test this change so this is mostly a best effort change that hopefully someone from domino can work through to add in the way that you all want. for context, we used [hardcoded values](https://github.com/braze-inc/cdk-cf-eks/tree/philnielsen/hardcode-service-ip-range) to spin up our cluster from our fork.

Let me know if you have any questions!


philnielsen:
closing as we are on TF now and I don't want this in my PR feed, feel free to reopen if you want to

PR #122, opened 2022-11-07
Use schema 1.2

fraenkel:


domino-pablo-rodriguez (approved):
lgtm

cdk/domino_cdk/config/template.py
```diff
@@ -187,6 +188,8 @@ def add_nodegroups(
                 }
             },
         )
+    if domino_version:
+        overrides["version"] = domino_version
```
Secretions, line 192: Dev overrides aside, since we're taking this as an arg rather than relying on a manually-provided override for the version, the most proper thing would be to add it to the install config object and pass it through to here:

https://github.com/dominodatalab/cdk-cf-eks/blob/a8fe918da4941c15baba1867e14f10a72c401708/cdk/domino_cdk/agent.py#L25

Secretions (approved):
Won't block on the thing I brought up since I think maybe next release or so we stop doing this entirely anyway, but this is a bit of a mixture of approaches so I still wanted to bring it up.

cdk/domino_cdk/config/template.py
fraenkel: technically I can see if just passing the schema and domino_version over as an override is sufficient.

fraenkel: right so this is where things can go sideways. as an override it lives in your deploy.yaml. so if you regenerate your terraform, it could change the domino version if we aren't careful.
The second issue is what we have documented as steps.
https://docs.dominodatalab.com/en/latest/admin_guide/e0f2ff/provision-infrastructure-and-runtime-environment/

fraenkel: So I see in the documented steps we have them create the config from the template and provide a domino version. I believe that is what we are missing on the CI and in platform-apps when we run through the deployer.

fraenkel: This is now all removed other than setting the schema.
When one matches on schema, the config is merged and we could get away with a subset of config. However, if this generated config is used in a future version that migrates, it would break so for now its generating a complete config.

PR #123, opened 2022-11-07
prefer default addon versions over latest

steved:
Updating the vpc-cni addon up to 1.11 (which is the latest) fails for clusters still on 1.9 as it cannot be jumped more than 1 version. If we prefer the default (which is 1.10) for now, we buy ourselves some time.

See #118 for the original change.


fraenkel (approved):
Given no one updates their CDK, won't this be a problem once 1.11 is the default?

PR #125, opened 2022-12-07
PLAT-6040: Use partition in all arns

fraenkel:
To support govcloud, the correct partition must be used when building ARNs.


fraenkel:
Verified locally that I could deploy to our govcloud account.

Secretions (approved):

PR #126, opened 2023-01-04
PLAT-5668: Conversion framework for CDK -> Terraform

Secretions:
This provides a process to convert EKS state from CDK to terraform.

It consists of the following:

* `README.md` - Instructions
* `terraform/*` - Master terraform wrapper around module
* `cloudformation-only/*` - Terraform to create special IAM role for deleting cloudformation stacks w/o deleting resources
* `migrate.py` - Migration script

The overall process is:

* `./migrate.py create-tfvars` - Create variables for terraform
* `./migrate.py get-imports` - Create script to import resources into terraform
* `terraform init`
* Run shell script generated by `get-imports`
* `terraform plan/apply`
* Scale down old ASGs
  * Domino will be down until it comes up on new ASG
* `./migrate.py clean-stack` - Delete unneeded cdk-only resources
* Do `terraform apply` in `cloudformation-only` subdirectory to create IAM role
* `./migrate.py delete-stack` - Delete the stack w/o deleting resources
* DONE!


convert/README.md
```diff
@@ -0,0 +1,185 @@
+# Domino CDK to Terraform migration
+
+Domino is deprecating this CDK repository, and switching all future EKS
+provisioning work to our Terraform module, located here:
+
+https://github.com/dominodatalab/terraform-aws-eks
+
+The scripts in this directory will help you convert the management your
```
ddl-dclegg, line 8: I think this will be clearer without `the management`

Secretions: The reason I included that phrase is to try and reinforce the idea that this isn't any sort of lift-and-shift migration.

convert/lib/convert.py
```diff
@@ -0,0 +1,614 @@
...
+
+        if not no_stacks:
+            self.sanity()
+            self.stacks = self.get_stacks(full=full)
+
+    def sanity(self):
+        extra_args = {}
+        stacks = []
+        while True:
+            list_result = self.cf.list_stacks(
```
steved, line 67: Take a look at the paginator methods: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#paginators

Here's an example in our code: https://github.com/cerebrotech/platform-apps/blob/16db2a357e0c647c05e53efb63ae29f7df885f14/src/services/cloud_providers/aws/resources.py#L1075

convert/lib/convert.py
```diff
@@ -0,0 +1,614 @@
...
+
+
+class app:
+    def __init__(self):
+        self.parse_args()
+        self.args.command()
+
+    def get_stacks(self, stack: str = None, full: bool = False):
+        stacks = {"resources": {}}
+        for s in self.cf.describe_stack_resources(StackName=stack or self.stack_name)[
```
steved, line 34: This doesn't need to be paginated?

convert/lib/convert.py
```diff
@@ -0,0 +1,614 @@
...
+        else:
+            resource_map = self.generate_resource_map(
+                availability_zones=self.args.availability_zones
+                or cdkconfig["vpc"]["max_azs"],
+                efs_backups=cdkconfig["efs"]["backup"]["enable"],
+                route53=bool(cdkconfig["route53"]["zone_ids"]),
+            )
+
+        def t(val: str) -> str:
+            val = re.sub(r"{{region}}", self.region, val)
```
steved, line 318: You can actually use the built-in format function for this exactly if you only use one bracket:
```
>>> "{stack}".format(stack="hello")
'hello'
```

convert/lib/nuke.py
```diff
@@ -0,0 +1,434 @@
...
+from .meta import cdk_ids
+
+
+class nuke:
+    def __init__(self, region: str, verbose: bool = False, delete: bool = False):
+        self.region = region
+        self.verbose = verbose
+        self.delete = delete
+
+    @property
```
steved, line 15: `@cached_property` (in functools) might work here?

convert/lib/nuke.py
```diff
@@ -0,0 +1,434 @@
...
+            self._ssm = boto3.client("ssm", self.region)
+        return self._ssm
+
+    @property
+    def stepfunctions(self):
+        if not getattr(self, "_stepfunctions", None):
+            self._stepfunctions = boto3.client("stepfunctions", self.region)
+        return self._stepfunctions
+
+    def _aws_list_all(
```
steved, line 51: Definitely see if the paginators help here?

convert/lib/nuke.py
```diff
@@ -0,0 +1,434 @@
...
+    def instance_profile(self, instance_profiles: list[str]):
+        if not instance_profiles:
+            return
+        existing_profiles = self._get_from_paged_list(
+            instance_profiles,
+            self.iam.list_instance_profiles,
+            lambda x: [p["InstanceProfileName"] for p in x["InstanceProfiles"]],
+        )
+
+        if existing_profiles:
```
steved, line 207: Pretty nitpicky, but I like this pattern since it keeps code from extending rightwards:

```
if not existing_profiles:
  return

pprint()

if not self.delete:
  return

...
```

convert/lib/nuke.py
```diff
@@ -0,0 +1,434 @@
...
+                try:
+                    layer_result = self.awslambda.get_layer_version(
+                        LayerName=layer, VersionNumber=version
+                    )
+                    if layer_result["LayerVersionArn"] == layerversion_arn:
+                        layerversion_results[layerversion_arn] = {
+                            "LayerName": layer,
+                            "VersionNumber": version,
+                        }
+                except:
```
steved, line 331: Can / should this be a specific type of exception? Is it a 404 essentially?

convert/lib/nuke.py
```diff
@@ -0,0 +1,434 @@
...
+    ):
+        all_referenced_groups = {}
+        if security_groups := nuke_queue.get(cdk_ids.security_group.value):
+            for sg in security_groups:
+                if referenced_groups := [
+                    r
+                    for r in self.ec2.describe_security_group_rules()[
+                        "SecurityGroupRules"
+                    ]
+                    if r.get("ReferencedGroupInfo", {}).get("GroupId") == sg
```
steved, line 371: I think `describe_security_group_rules` takes a specific Group ID as a filter in the API call?

convert/lib/convert.py
```diff
@@ -0,0 +1,614 @@
...
+                    if logical_id.startswith(mapped_logical_id):
+                        stacks[name] = self.get_stacks(physical_id, full)
+                        break
+                else:
+                    raise Exception(f"Nothing to map stack {s} to!")
+            else:
+                if re.match("^[0-9A-F]{8}", logical_id[-8:]):
+                    stacks["resources"][logical_id[:-8]] = s if full else physical_id
+                else:
+                    stacks["resources"][logical_id] = s if full else physical_id
```
steved, line 50: just a bit cleaner IMO

```suggestion
                key = local-id[:-9] if re.match("^[0-9A-F]{8}", logical_id[-8:]) else logical_id
                stacks["resources"][key] = s if full else physical_id
```

convert/lib/nuke.py
Secretions: This is seeing if the group id in the `ReferencedGroupInfo` part of the response is the same. We're specifically looking for groups rules for groups other than the one in question, that have a rule with a source of the group in question.

convert/lib/nuke.py
Secretions: It doesn't go very far right, while that pattern would add four lines to every type of thing we nuke.

convert/lib/convert.py
Secretions: For some reason I had it in my head that boto3's paginators were part of the resource api and I didn't bother to look at them. Will try them out.

Secretions: Switched to the paginators across the board.

convert/lib/convert.py
Secretions: Our stacks appear to have fewer than 100 resources, but I did mean to get back to this later. I've paginated it just in case.

convert/lib/nuke.py
Secretions: Yep, quite a bit cleaner now!

convert/lib/convert.py
Secretions: This made it occur to me that, since we have a couple actual formatted input strings that this is in causing us to have to do `{{{{stack_name}}}}`, that probably the best thing to do is keep the current substitution method and change the sigil away from braces altogether. Four braces is madness, and two into a format gives me vietnam flashbacks of our deployer jinja turduckens.

I made it a `%`. Since we use `re` elsewhere, I'm leaving it as `re.sub` (as opposed to maybe doing `"".replace`, say).

convert/README.md
```diff
@@ -0,0 +1,185 @@
...
+Note that the ssh key is handled different in the terraform module. With CDK,
+we used an existing keypair, while the terraform module creates a keypair for
+you. The local key does not need to be the one you used with CDK. The new
+autoscaling group nodes and bastion node will be provisioned with the new key.
+
+### Get AWS resources to import
+
+Then, we'll use `convert.py` to generate the Terraform import commands to import our AWS resources:
+
+    ./convert.py get-imports --region us-east-1 --stack-name mystack > terraform/imports.sh
```
ddl-dclegg, line 70: We need to add `chmod +x terraform/imports.sh` here or prior to executing it below.

ddl-dclegg (changes requested):

convert/README.md
```diff
@@ -0,0 +1,185 @@
...
+	delete-stack        Get commands to delete old stack
+	print-stack         Print CDK stack resources
+
+### Create terraform variables
+
+First, we must create the variables going into our Terraform config.
+
+The following command will look at the current CDK stack and autopropagate the Terraform variables:
+
+    ./convert.py create-tfvars --region us-east-1 --stack-name mystack --ssh-key-path /path/to/key.pem > terraform/terraform.tfvars
```
ddl-dclegg, line 59: This needs to output to `terraform.tfvars.json` instead of `terraform.tfvars`. The content is valid json, not an HCL or key-value format, so when it is used later in `imports.sh` it throws an error.

```
+ ./imports.sh
+ terraform import module.domino_eks.module.storage.aws_efs_file_system.eks fs-0172b729c2ee51a56
╷
│ Error: Argument or block definition required
│
│   on terraform.tfvars line 1:
│    1: {"deploy_id": ...
```
Using `tfvars.json` clears that issue.

ddl-dclegg (changes requested):

convert/README.md
Secretions: Doh, yeah. I've made that mistake working with it manually as well, so not suprising... fixed.

convert/lib/convert.py
```diff
@@ -0,0 +1,545 @@
...
+                "(LogRetention|backuppostcreationtasks).*": lambda_safe,
+            },
+            "eks_cluster_stack": {
+                "(LogRetention|IsCompleteHandler|NodeProxyAgentLayer|OnEventHandler|Provider).*": lambda_safe,
+            },
+            "eks_kubectl_stack": {
+                "(Handler|KubectlLayer|ProviderframeworkonEvent).*": lambda_safe,
+            },
+            "eks_stack": {
+                f"(snapshot|chibipug7501ebscsi|{self.stack_name}DominoEcrRestricted|autoscaler)": [
```
ddl-dclegg, line 394: Not sure if this will cause an issue, but I just noticed `chibipug7501ebscsi` here

Secretions: Yep, will fix this. Should have the stack name instead.

convert/README.md
Secretions: So the directions here actually do say to just run `bash imports.sh`, so the chmod isn't actually necessary as the instructions are written. I can totally understand reading the instructions and abstracting it into the way you're habituated to operating, though. I could add the chmod in case that happens to anyone else, but on the other hand it adds an extra step. Not sure which is the better option.

ddl-dclegg: Valid arguments. I'm okay with moving forward.

ddl-dclegg (approved):
I think I'm good with this, provided @steved has no lingering/further objections.

ddl-dclegg:
> I think I'm good with this, provided @steved has no lingering/further objections.

FWIW, I used this manually, and then scripted with a [dominator helper](https://github.com/cerebrotech/qe-dominator/blob/4e4380c75479354a50a290d3bd8e30668374c5cf/utils/cdkConvert.sh) to convert `dclegg7933` (a 5.1.3 cdk deployment).

convert/lib/convert.py
```diff
@@ -0,0 +1,545 @@
...
+        ]
+        for r in resources:
+            logical_id = r["LogicalResourceId"]
+            physical_id = r["PhysicalResourceId"]
+            if r["ResourceType"] == cdk_ids.cloudformation_stack.value:
+                for mapped_logical_id, name in stack_map.items():
+                    if logical_id.startswith(mapped_logical_id):
+                        stacks[name] = self.get_stacks(physical_id, full)
+                        break
+                else:
```
steved, line 46: TIL for-else...

convert/lib/convert.py
```diff
@@ -0,0 +1,545 @@
...
+
+    def create_tfvars(self):
+        self.setup()
+        cdkconfig = yaml.safe_load(
+            [
+                o["OutputValue"]
+                for o in self.cf.describe_stacks(StackName=self.stack_name)["Stacks"][0]["Outputs"]
+                if o["OutputKey"] == "cdkconfig"
+            ][0]
+        )
```
steved, line 311: Might be worth plucking this out since it's used in get_imports too?

convert/lib/convert.py
```diff
@@ -0,0 +1,545 @@
...
+            "region": cdkconfig["aws_region"],
+            "grandfathered_creation_role": self.stacks["eks_stack"]["resources"]["eksCreationRole"],
+            "tags": {**cdkconfig["tags"], "domino-deploy-id": cdkconfig["name"]},
+            "vpc_id": cdkconfig["vpc"]["id"]
+            if not cdkconfig["vpc"]["create"]
+            else self.stacks["vpc_stack"]["resources"]["VPC"],
+            "public_subnet_ids": get_subnet_ids("Public"),
+            "private_subnet_ids": get_subnet_ids("Private"),
+            "pod_subnet_ids": get_subnet_ids("Pod", ""),
+            "k8s_version": cdkconfig["eks"]["version"],  # We're trusting this is accurate
```
steved, line 346: Is it better to just check & use the EKS cluster data?

convert/lib/convert.py
```diff
@@ -0,0 +1,545 @@
...
+            "ssh_key_path": self.args.ssh_key_path,
+            "number_of_azs": cdkconfig["vpc"]["max_azs"],
+            "route53_hosted_zone_name": reduce(
+                lambda cfg, k: cfg[k] if k in cfg else [""],
+                ["install", "overrides", "external_dns", "domain_filters"],
+                cdkconfig,
+            )[0],
```
steved, line 353: Could we just us the zone ID from the config and map it to the name in `main.tf`?

convert/terraform/main.tf
```diff
@@ -0,0 +1,44 @@
...
+
+module "domino_eks" {
+  source                       = "github.com/dominodatalab/terraform-aws-eks.git?ref=smooth-conversion-process"
+  deploy_id                    = var.deploy_id
+  region                       = var.region
+  number_of_azs                = var.number_of_azs
+  k8s_version                  = var.k8s_version
+  route53_hosted_zone_name     = var.route53_hosted_zone_name
+  eks_master_role_names        = var.eks_master_role_names
+  s3_force_destroy_on_deletion = true
```
steved, line 34: I guess this doesn't _really_ matter, but should this also come from the CDK config?

convert/lib/convert.py
```diff
@@ -0,0 +1,545 @@
...
+            },
+            "vpc_stack": {
+                "endpointssg": [cdk_ids.security_group.value],
+                "bastion": [
+                    cdk_ids.instance.value,
+                    cdk_ids.instance_profile.value,
+                    cdk_ids.iam_role.value,
+                    cdk_ids.eip.value,
+                ],
+                "VPCrejectFlowLogsFlowLog": [cdk_ids.flowlog.value],  # TODO: should this be in the module?
```
steved, line 413: FWIW to answer the question, I think yes? Why not import it?

convert/cloudformation-only/main.tf
```diff
@@ -0,0 +1,40 @@
+resource "aws_iam_policy" "cloudformation_only" {
```
steved, line 1: Maybe it's overkill, but I feel like just creating & deleting this role in Python when `delete-stack` is called is a nicer "experience", but it's not a huge deal. It's also not a unique name, but that is only a problem for testing.

convert/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Delete the old cloudformation stack
+
+Enter the `cloudformation-only/` subdirectory, and provision that terraform:
+
+    cat EOF >> terraform.tfvars
+    region="us-east-1"
+    tags={}
+    EOF
+    terraform plan -out=terraform.plan
```
steved, line 159: TF init too

convert/terraform/variables.tf
```diff
@@ -0,0 +1,105 @@
...
+}
+
+variable "ssh_key_path" {
+  type        = string
+  description = "Path to private SSH key"
+}
+
+variable "number_of_azs" {
+  type        = number
+  description = "Number of AZs used in deployment (usually 3)"
```
steved, line 56: I wonder if it's a bit more descriptive as # of subnets given the convo today? Though I can come back through here and clean it up once the other changes land

convert/lib/nuke.py
```diff
@@ -0,0 +1,351 @@
...
+            pprint({"Auto scaling groups to delete": existing_groups})
+
+            if self.delete:
+                for group in existing_groups:
+                    print(self.autoscaling.delete_auto_scaling_group(AutoScalingGroupName=group))
+
+    def eip(self, eip_addresses: list[str]):
+        if not eip_addresses:
+            return
+        result = self.ec2.describe_addresses()
```
steved, line 62: Doesn't need to be paginated?

steved (approved):

convert/lib/nuke.py
Secretions: I'll double check, but I'm pretty sure this was one of the things where the api actually returns everything.

convert/terraform/variables.tf
Secretions: Makes sense

convert/cloudformation-only/main.tf
Secretions: I kind of figured that once we're creating things, it's better to have terraform with state etc. Additionally, I also made this something that can be done manually for the paranoid, so having that managed outside of the script makes sense for that.

convert/lib/convert.py
Secretions: I'll make a ticket to add the functionality to terraform-aws-eks. I'll keep this here for the sake of tracking and update it in tandem.

convert/terraform/main.tf
Secretions: Oops, yes, this was an oversight.

convert/lib/convert.py
Secretions: I was probably trying to avoid making aws api calls, but since we're ostensibly getting rid of the config pass through bit this probably needs to happen.

convert/lib/convert.py
Secretions: I'm sure in part I was trying to avoid extra api calls where unnecessary, but I also genuinely am not sure. If there's a mismatch, we could handle that situation in a number of ways (silently do what we think would work, silently do what the config is as above and let terraform be where it's discovered, or do some "mismatch" error with an option to bypass).

convert/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[tool.black]
```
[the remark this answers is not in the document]
Secretions: I'll make a requirements. I think this basically just requires boto3 and pyyaml, maybe one or two other dumb standard-ish things.

convert/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+The *old* ones look like this:
+
+    clustername-platform-0-us-east-1
+
+The *new* ones look like this:
+
+    eks-clustername-platform-0-abcd1234-abcd-1234-5678-abcd12345678
+
+Once all the Domino pods are live on the *new* nodes, Domino should be back up and functional.
```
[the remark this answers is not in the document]
Secretions: Part if what's shown in the readme is out of date. I made it so the ASGs actually work right after provision. I'll update that top part of the phrasing.

In a follow up thing, I'm adding the ability to delete the ASGs. This will be without a kubectl cordon/drain, so ungraceful. My plan is to make it so this can still be semi-automated via clean-stack doing this dumber procedure. I'd prefer to avoid automating anything on the kubectl level other than what's already done already in the terraform-aws-eks.

convert/lib/nuke.py
Secretions: Yeah, there's no paging for this call.

convert/terraform/variables.tf
Secretions: Actually, I'm going to leave it as-is for now since it still maps directly onto the module's variable which the same variable name and similar description wording.

convert/terraform/main.tf
Secretions: Also does matter because we want this off in the real world. :)

Secretions: There is some slight ambiguity in how to handle this, since the config doesn't directly map, so I'll add this to the follow up PR.

convert/README.md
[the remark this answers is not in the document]
Secretions: Updated the top part to be accurate. Since I'll be working on the ASGs anyway, if we do want to finesse it beyond that can address in follow up (and work with a smaller changeset...).

Secretions:
Leaving a few minor things in review for followup PR, see https://github.com/dominodatalab/cdk-cf-eks/pull/131

PR #127, opened 2023-01-09
PLAT-6117: Upgrade CDK for EKS 1.24

fraenkel:
CDK includes https://github.com/aws/aws-cdk/pull/22945


steved (approved):
Update the [default](https://github.com/dominodatalab/cdk-cf-eks/blob/80101064a8e22f055c4aee4e78fd4bc9f1fc2abb/cdk/domino_cdk/config/template.py#L133) too?

PR #129, opened 2023-01-31
remove agent_template.yaml from terraform generation

steved:
Agent template handling has proven difficult and at odds with how we generate it for other clouds/Terraform.


fraenkel:
Does upgrading from an older CDK work to this version?

fraenkel (approved):

cdk/domino_cdk/domino_stack.py
```diff
@@ -147,24 +146,4 @@ def generate_outputs(self):
                 value=r53_owner_id,
             )
 
-        if self.cfg.install is not None:
-            agent_cfg = generate_install_config(
```
Secretions, line 151: If we outright remove the template, we should also rewrite this:

https://docs.dominodatalab.com/en/latest/admin_guide/c1eec3/deploy-domino/#_configure_the_cluster

Secretions (commented):
We need to make a decision between:

* Keep the agent template, but deployer just doesn't use it
  * Can continue testing Domino
  * CDK-only people have the same workflow
  * Schema might upgrade cleanly coming just from here, for a while, but...
  * We'll still have to futz with it, so don't get the full goal
* Remove the template generation, but continue to test domino
  * Need to find a way to generate the agent config in the repo
  * Need to rewrite the install documentation to account for this
* Remove the template generation, only test deployment
  * Need to rewrite the install documentation
  * No longer testing Domino (but we don't do that elsewhere?)
  * Don't need to care about version/schema changes anymore

Secretions (approved):

PR #130, opened 2023-02-03
PLAT-6183: Update gpu remediation

Secretions:
They release the fixed GPU ami, however, it was not, in reality, fixed.

See: https://github.com/awslabs/amazon-eks-ami/issues/1154#issuecomment-1414711705

Because the incorrect version of the certs config triggers the safety valve of the old remediation, a new one is needed.


fraenkel (approved):
its getting cleaner

PR #131, opened 2023-02-17
PLAT-5668: TF convertor integration test / misc. follow-up

Secretions:
DONE:
* Automatically destroy ASG instances before ASGs
* Deal with optionality of bastion
* Same with monitoring bucket
* Stack name is sanitized to `\w` or something similar, deal with that
* Public API setting unsupported, notify user
* Don't hardcode S3 force destroy variable
* Get route53 zone name from api; agent config going away
* Maybe get EKS version from api?
* Support flowlog import into eks module if/when added to that
* Determine if we should do something more clever for the ASG node draining
* Optionality between managed/unmanaged nodegroups, efs backups, comb for other things
* Track the upstream AZ/Subnet changes and change input variable accordingly (if relevant)
* (barely) Manage the default auto-generated EKS SG because its destruction is unreliable
* Integration test

Still TODO:
* _Will couple deployer support with this, so any further changes needed on this side for that_
  * Nevermind, I'll start this in platform-apps after this merges and make a new PR if any changes are still required. This ended up being more of a thing than I anticipated...


Secretions:
See this for full run: https://github.com/dominodatalab/cdk-cf-eks/actions/runs/4299341793/jobs/7494451644

.github/workflows/test.yml
```diff
@@ -15,6 +15,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       DEPLOYER_IMAGE: quay.io/domino/deployer:develop.f72b81d5db7e04cc48478d310eafa4abb927ce7f
```
steved, line 17: I missed this originally, it's no longer used and the associated step is no longer useful yeah?

convert/cloudformation-only/main.tf
```diff
@@ -14,7 +14,7 @@ resource "aws_iam_policy" "cloudformation_only" {
 }
 
 resource "aws_iam_role" "cloudformation_only" {
-  name = "cloudformation-only"
+  name = "cloudformation-only${var.suffix}"
```
steved, line 17: Both of these resources support `name_prefix` if you want TF to handle it

.github/workflows/test.yml
```diff
@@ -100,28 +97,86 @@ jobs:
...
+        ./convert.py print-stack --stack-name $NAME --region $AWS_REGION --verbose --yaml > stack-data.yaml
+        ./convert.py create-tfvars --stack-name $NAME --region $AWS_REGION --ssh-key-path ./dummy.pem > terraform/terraform.tfvars.json
+        ./convert.py get-imports --stack-name $NAME --region $AWS_REGION > terraform/imports.sh
+    - name: Run terraform
+      if: (contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master')
+      working-directory: ./convert/terraform
+      run: |
+        terraform init
+        bash imports.sh
+        terraform plan -out=terraform.plan
```
steved, line 150: Wonder if it's worth checking for resource destruction (vs. add/change) and failing in here if there are any?

steved (approved):
LGTM

.github/workflows/test.yml
Secretions: Yeah, I nuked it it.

.github/workflows/test.yml
Secretions: Good call, I added a thing.

convert/cloudformation-only/main.tf
Secretions: Doing it this way means I can also associate it with the deploy name.

PR #133, opened 2023-03-09
PLAT-6357 Install calico using helm chart

miguelhar:
[PLAT-6357](https://dominodatalab.atlassian.net/browse/PLAT-6357)

We are currently using the k8s manifest to install the calico operator which are hosted in the aws repo. As of 1.25 aws is no longer hosting the calico-operator.yaml .

https://docs.aws.amazon.com/eks/latest/userguide/calico.html 

We need to update the cdk automation to account for these changes.

Updating the current calico installation to use the [tigera-operator](https://raw.githubusercontent.com/projectcalico/calico/v3.25.0/manifests/tigera-operator.yaml) manifests caused an error while installing.
```
357330 byte payload is too large for the Event invocation type (limit 262144 bytes) (Service: AWSLambda; Status Code: 413; Error Code: RequestEntityTooLargeException; Request ID: 28b8e09f-026f-43c5-8dbb-754bcfcd4af5; Proxy: null)
```
This is by a single k8s object: 
```
name: installations.operator.tigera.io
size: 347.61 KB.
```
As a result we are updating the [calico-installation](https://github.com/dominodatalab/cdk-cf-eks/blob/master/cdk/domino_cdk/aws_configurator.py), to use the helm-chart approach, which ultimately seems to be recommended.


Secretions (approved):
Was thinking this would have to wait for the helm change on the other side, but maybe the other being pre-operator was the only blocker?

miguelhar:
> 

I think this doesn't have to wait for the module, since we are most likely migrating from earlier versions that did not use helm. The migration however is dependant of the terraform module ability to delete then install, which Im working  on atm

miguelhar:
@Secretions @steved @fraenkel

miguelhar:
> > 
> 
> I think this doesn't have to wait for the module, since we are most likely migrating from earlier versions that did not use helm. The migration however is dependant of the terraform module ability to delete then install, which Im working on atm

It actually does, since we need to update the `convert` with a module that is able to upgrade calico

cdk/domino_cdk/aws_configurator.py
```diff
@@ -32,53 +10,17 @@ def __init__(self, scope: cdk.Construct, eks_cluster: eks.Cluster):
...
-            )
-
-        if notcrd_manifests:
-            non_crds = eks.KubernetesManifest(
-                self.scope, "calico", cluster=self.eks_cluster, manifest=notcrd_manifests, overwrite=True
-            )
-
-            for crd in crds:
-                non_crds.node.add_dependency(crd)
+        eks.HelmChart(
```
steved, line 13: Does this work ok for upgrades? You don't need the same deletion as the TF side?

miguelhar: I dont think it can as CF recreates the EKS stack.

miguelhar: I was wrong, cdk attempts to upgrade but fails ti apply the helm release since it calico is installed. I thought CF would remove the resources it created.

miguelhar:
Wont do

PR #134, opened 2023-03-10
PLAT-6279: CDK->Terraform conversion process changes for deployer compatibility

Secretions:
Final round of initial changes for this process, the things needed for a converted deployer deploy to be able to be managed normally after conversion:

* Use the same name for the module
* Move non-module things outside of main
* Move deployer outputs *into* main
* Disconnect EKS module from flow bucket
* Match a few config variables that were missing
* Change main.tf to main.tf.json so the deployer can generate on top of it.


convert/terraform/outputs.tf
```diff
@@ -1,49 +1,4 @@
 output "domino_eks" {
```
steved, line 1: is this output used anymore?

convert/terraform/variables.tf
```diff
@@ -116,17 +116,47 @@ variable "eks_master_role_names" {
...
+
+variable "kubeconfig_path" {
+  type        = string
+  description = "fully qualified path name to write the kubeconfig file"
+  default     = ""
+}
+
+variable "flow_log_bucket_arn" {
+  type        = string
+  description = ""
```
steved, line 160: Fill this in or remove it before merging?

steved (approved):

convert/terraform/outputs.tf
Secretions: Was debating whether I wanted to keep it or not. The other outputs are specifically for the deployer, but this here (alone, without those outputs) is what I would probably do if I were making something that was only ever used without the deployer. As this is used with both, but has no obligation to track the deployer outputs as-is (since the deployer will now overwrite main.tf.json), there's maybe still both user-friendliness/completeness arguments for this for any usage of this that doesn't coincide with the deployer.

convert/terraform/variables.tf
Secretions: Doh, updated.

PR #137, opened 2023-04-04
PLAT-6496 Update tf-aws-eks mod var schema

miguelhar:
[PLAT-6496](https://dominodatalab.atlassian.net/browse/PLAT-6496)

### DIFF
--- a/cdk/setup.py
+++ b/cdk/setup.py
@@ -38,7 +38,7 @@ setuptools.setup(
         "aws-cdk.core~=1.187.0",
         "aws-cdk.lambda-layer-awscli~=1.187.0",
         "aws-cdk.lambda-layer-kubectl~=1.187.0",
-        "boto3~=1.26.0",
+        "boto3==1.26.90",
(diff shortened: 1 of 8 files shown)


cdk/setup.py
```diff
@@ -38,7 +38,7 @@
         "aws-cdk.core~=1.187.0",
         "aws-cdk.lambda-layer-awscli~=1.187.0",
         "aws-cdk.lambda-layer-kubectl~=1.187.0",
-        "boto3~=1.26.0",
+        "boto3==1.26.90",
```
miguelhar, line 41: Updated to fix https://github.com/boto/boto3/issues/3648

Secretions: You can rebase and revert this. I ended up fixing this in a different way that I think is slightly more correct.

Secretions (approved):
lgtm other than the boto thing

cdk/setup.py
miguelhar: done

PR #140, opened 2023-04-18
PLAT-6563: monitoring s3 bucket object ownership

fraenkel:
Set ownership to BucketOwnerPreferred.


fraenkel:
Verified with locally provisioned cdk and then converted to eks.

PR #142, opened 2023-04-24
PLAT-6503: Support converting customer-provided CDK KMS ids

Secretions:
If customer-provided KMS keys were used, migrate that configuration correctly.

It's worth noting that the way the CDK and terraform modules are respectively configured is not perfectly compatible. If a single KMS key was used across-the-board on both S3 buckets and the EKS control plane, then we will use it. Otherwise, we will turn it on, enter a non-validating dummy value, and inform the user to pick a key.

One extra nicety I *didn't* add is just going with a key should a single key be in the configuration, but not used with all services. I can imagine in this configuration, the one a user would want to migrate to is one using that key for everything, so perhaps I should configure it to work this way? Unsure.


convert/lib/convert.py
```diff
@@ -443,12 +454,25 @@ def get_subnet_ids(subnet_type: str, prefix: str = "VPC"):
             "eks_cluster_auto_sg": eks_cluster_auto_sg,
         }
 
+        sse_kms_key_ids = [v["sse_kms_key_id"] for k, v in self.cdkconfig["s3"]["buckets"].items()]
+        sse_kms_key_ids.append(self.cdkconfig["eks"]["secrets_encryption_key_arn"])
+        kms_id_note = False
+        if not all(True for key_id in sse_kms_key_ids if key_id is None):
```
steved, line 460: just a suggestion, it's clearer in my head

```suggestion
        if any(key_id is not None for key_id in sse_kms_key_ids):
```

steved (approved):
> One extra nicety I didn't add is just going with a key should a single key be in the configuration, but not used with all services. I can imagine in this configuration, the one a user would want to migrate to is one using that key for everything, so perhaps I should configure it to work this way? Unsure.

Is it even possible to migrate like this? I know it's a PITA to swap keys for EKS, not sure about S3/EFS/etc.

> It's worth noting that the way the CDK and terraform modules are respectively configured is not perfectly compatible. If a single KMS key was used across-the-board on both S3 buckets and the EKS control plane, then we will use it. Otherwise, we will turn it on, enter a non-validating dummy value, and inform the user to pick a key.

+1, we could explode the module (or allow overrides) if it's truly necessary, but I'm skeptical KMS keys are used anywhere

Secretions:
> Is it even possible to migrate like this? I know it's a PITA to swap keys for EKS, not sure about S3/EFS/etc.

I'm going to test both the ideal and not-deal KMS scenarios (eks-but-not-s3 and s3-but-not-eks) before I merge and note the expected situation or tune this approach depending on the outcome.

> I'm skeptical KMS keys are used anywhere

While I wouldn't be totally surprised if it is totally unused, I think the S3 kms ids being used in an install or three is pretty plausible. My instinct is that if anybody has used any variation of the kms options, it's likely to be either all s3 buckets using a single kms id but not setting the eks kms option, or that but without the monitoring bucket (or with monitoring outright turned off).

Secretions:
SOOOoooo..... turns out there is a bug in loading the KMS key id config for S3 buckets. Given the extremely low likelihood that someone wanted to override the KMS key for the EKS cluster itself, but not S3, I'm going to conclude the custom KMS feature was never utilized and close this.

PR #143, opened 2023-08-19
PLAT-7108: import EIP using eipalloc

miguelhar:
[PLAT-7108](https://dominodatalab.atlassian.net/browse/PLAT-7108)
cdk to terraform convert : terraform imports fail with 
```
Error: with the retirement of EC2-Classic standard domain EC2 EIPs are no longer supported
```

This is due to an upstream change in the terraform `aws` provider.
The fix is to import the `eip` resource using its allocation id rather than its public ip.

Misc Changes:
* `PRIVATE_WITH_NAT` `PRIVATE` Subnet definition deprecated.
* `JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION` reduce noise.
* Forcing the main stack into `DELETE_FAILED` used to cascade to nested stacks. Fix by iterating through nested stacks.
* Adds `ForceDelete` to ASG to mitigate:
```
botocore.errorfactory.ScalingActivityInProgressFault: An error occurred (ScalingActivityInProgress) when calling the DeleteAutoScalingGroup operation: You cannot delete an AutoScalingGroup while there are scaling activities in progress for that group.
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,13 +73,20 @@ jobs:
         REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
         ACM_CERT_ARN: ${{ secrets.DELTA_ACM_CERT_ARN }}
         BASE_DOMAIN: ${{ secrets.DELTA_BASE_DOMAIN }}
+        JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: 1
       run: |
         export NAME=cdk-${GITHUB_SHA:0:6}-$(date +%s)
         echo "NAME=$NAME" >> $GITHUB_ENV
         ./util.py generate_config_template --name $NAME --aws-region=$AWS_REGION --aws-account-id=$AWS_ACCOUNT_ID --dev --platform-nodegroups 2 --registry-username $REGISTRY_USERNAME --registry-password $REGISTRY_PASSWORD --hostname $NAME.$BASE_DOMAIN --acm-cert-arn $ACM_CERT_ARN --disable-flow-logs > config.yaml
         ./util.py load_config -f ./config.yaml
     - name: Test default config (single and nested stacks)
+      env:
+        JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: 1
       run: |
+        echo 'CDK acknowledge: AWS CDK v1 End-of-Support June 1, 2023'
+        cdk acknowledge 19836
+        echo 'CDK acknowledge: (eks) eks overly permissive trust policies'
+        cdk acknowledge 25674
         cdk synth --context singlestack=true -q
         cdk synth -q
     - name: Upload distribution artifacts
@@ -95,11 +102,18 @@ jobs:
         urlfile=$(python -c 'import sys, urllib.parse; print(urllib.parse.quote(sys.stdin.read().strip()))' <<< "$filename")
         echo "Artifact url: https://domino-artifacts.s3.amazonaws.com/cdk/$($DATEDIR)/$urlfile"
         done
+    - name: Bootstrap CDK
+      env:
+        AWS_ACCOUNT_ID: ${{ secrets.DELTA_ACCOUNT_ID }}
+        AWS_REGION: ${{ env.AWS_REGION }}
+        JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: 1
+      run: cdk bootstrap "aws://$AWS_ACCOUNT_ID/$AWS_REGION"
     - name: Deploy CDK
       if: contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master'
       env:
         REGISTRY_USERNAME: ${{ secrets.REGISTRY_USERNAME }}
         REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
+        JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: 1
       run: |
         docker login -u $REGISTRY_USERNAME -p $REGISTRY_PASSWORD quay.io
         cdk deploy --require-approval never --outputs-file outputs.json
@@ -187,15 +201,17 @@ jobs:
         aws-region: ${{ env.AWS_REGION }}
     - name: Delete stack w/CDK
       if: always() && (contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master')
+      env:
+        JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: 1
       working-directory: ./cdk
       run: |
         cdk destroy --force
-    - name: Destroy Infrastructure
+    - name: Destroy Infrastructure tf
       if: always() && (contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master')
       working-directory: ./convert/terraform
       run: |
         terraform destroy -auto-approve
-    - name: Destroy Infrastructure
+    - name: Destroy Infrastructure cf
       if: always() && (contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master')
       working-directory: ./convert/cloudformation-only
       run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -26,182 +26,209 @@ jobs:
...
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE }}
+          aws-region: ${{ env.AWS_REGION }}
+      - name: Delete stack w/CDK
+        if: always() && (contains(github.event.pull_request.labels.*.name, 'deploy-test') || github.ref == 'refs/heads/master')
+        env:
+          JSII_SILENCE_WARNING_DEPRECATED_NODE_VERSION: 1
+        working-directory: ./cdk
+        run: |
+          MAX_RETRIES=3
```
Secretions, line 202: I suspect this retry was added trying to troubleshoot problems with deletions, but also probably wasn't/won't be the actual solution. I'd rather not do a retry. If it fails, it fails.

cdk/domino_cdk/provisioners/efs.py
```diff
@@ -45,7 +51,7 @@ def provision_efs(self, stack_name: str, cfg: config.EFS, vpc: ec2.Vpc, security
             security_group=security_group,
             performance_mode=efs.PerformanceMode.GENERAL_PURPOSE,
             throughput_mode=efs.ThroughputMode.BURSTING,
-            vpc_subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PRIVATE),
+            vpc_subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PRIVATE_WITH_NAT),
```
Secretions, line 54: Did CDK get upgraded on us somehow? Like we failed to pin it correctly? I wouldn't expect there to be any required code changes here.

convert/lib/convert.py
```diff
@@ -223,6 +263,23 @@ def print_stack(self):
         else:
             pprint(out)
 
+    def _nested_az_replace(self, d: dict, count: int):
```
Secretions, line 266: Six of one half doesn't of the other to me, but curious why these inline functions were moved out? Since we don't have unit tests and it doesn't seem like we've changed either this or `get_stack_resources`, nor used them in any new places outside their old parent functions, I'm not sure what would inspire it.

convert/lib/convert.py
```diff
@@ -617,32 +690,23 @@ def delete_stack(self):
...
-                get_stack_resources(s)
+                self._get_stack_resources(s, stacks)
 
         for i, (stack, resources) in enumerate(stacks.items()):
             if self.args.delete:
-                self.cf.delete_stack(StackName=stack, RoleARN=cf_only_role["value"], RetainResources=resources)
+                if (
+                    stack_status := self.cf.describe_stacks(StackName=stack)["Stacks"][0]["StackStatus"]
+                ) and stack_status != "DELETE_FAILED":
+                    resources = None
```
Secretions, line 702: I'm still not 100% clear on what was actually going on with the CI failure (https://github.com/dominodatalab/cdk-cf-eks/actions/runs/5954998517/job/16152692447), and don't understand how the subsequent run with this code didn't exit with some failure. Note that it did leave all the nested stacks behind, so wasn't a true success, though I unfortunately destroyed the evidence as I cleaned them up to test that this stack deletion strategy still works in general).

Anyway, it sounds almost as if we got to this line of code without the stacks actually having a deletion attempt. We should not be able to get this point without the stacks being in `DELETE_FAILED`. The whole point of this workflow is that we intentionally delete the stacks in a way where they will all fail, which triggers the state where we can provide the list of resources. Makes me wonder if AWS changed it so the child stacks aren't ending up in `DELETE_FAILED`, and what we really need to do is iterate when we run it the first time.

We should probably also iterate and confirm that none of them are in non-`DELETE_FAILED` states, and report what those states are and exit with a failure, before we get to that second delete. Anything other than `DELETE_FAILED` (or `DELETE_COMPLETE` if you check before the line where we take those out) indicates that something went wrong.

convert/lib/convert.py
```diff
@@ -190,21 +218,33 @@ def parse_args(self):
         clean_stack_parser.set_defaults(command=self.clean_stack)
 
         delete_stack_parser = subparsers.add_parser(
-            name="delete-stack", help="Get commands to delete old stack", parents=[common_parser]
+            name="delete-stack",
+            help="Get commands to delete old stack",
+            parents=[common_parser],
```
Secretions, line 223: Unless our repo is doing it (in which case I'd think it that's a change somehow?), I'd prefer if you could set your black so it has a line length of 120 for these files. It'd avoid a lot of this noise. Don't mind anything that's actually poorly-formatted getting fixed, but I think the args here, at the very least, are just a black settings difference.

.github/workflows/test.yml
```diff
@@ -26,182 +26,209 @@ jobs:
...
-        black $f --check
-        isort $f --check
-        done
-    - name: Test with pytest
-      run: |
-        coverage run -m pytest tests
-    - name: Coverage report
-      run: |
-        coverage report
+      - uses: actions/checkout@v2
```
Secretions, line 29: Unless there's a part where the yaml is actually poorly formed, I'd rather not reformat all this. The dash positioning is arbitrary, while the diff obscures what the actual changes are.

Secretions (changes requested):
The actual EIP part looks good, but I think whatever was wrong with deletion wasn't addressed properly and want to be sure that the PRIVATE_WITH_NAT change doesn't come from the cdk version getting out of sync.

.github/workflows/test.yml
miguelhar: There are instances where the delete fail due to a sg, but I was able to delete it manually(hints at timing), given this is a delete, Id say its worth a retry rather than ~1hr for a re-run.

cdk/domino_cdk/provisioners/efs.py
miguelhar: API is being deprecated 
```
Warning:  @aws-cdk/aws-ec2.SubnetType#PRIVATE is deprecated.
  use `PRIVATE_WITH_NAT`
  This API will be removed in the next major release.
Warning:  @aws-cdk/aws-ec2.SubnetType#PRIVATE is deprecated.
  use `PRIVATE_WITH_NAT`
  This API will be removed in the next major release.
```

convert/lib/convert.py
miguelhar: just moved them to make `flake8` happy
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

convert/lib/convert.py:266:5: C901 'app.generate_resource_map' is too complex (17)
convert/lib/convert.py:639:5: C901 'app.delete_stack' is too complex (19)

```

cdk/domino_cdk/provisioners/efs.py
Secretions: I do remember seeing those at some point earlier, so wfm. Don't mind the change, just wanted to make sure it wasn't masking some other unintended change.

.github/workflows/test.yml
miguelhar: removed

convert/lib/convert.py
miguelhar: Initially misunderstood scope, Fixed by iterating through nested stacks

convert/lib/convert.py
miguelhar: fixed

ddl-dclegg (approved):
Looks good to me

ddl-dclegg:
@Secretions were your concerns resolved?

convert/lib/convert.py
```diff
@@ -586,6 +592,42 @@ def get_nukes(stack_name, stack_resources):
...
+        stack_details = self.cf.describe_stacks(StackName=stack_name)
+        stack_status = stack_details["Stacks"][0]["StackStatus"]
+
+        if stack_status != "DELETE_FAILED":
+            print(f"Deleting Stack: {stack_name} Status: {stack_status}")
+            self.cf.delete_stack(StackName=stack_name, RoleARN=role_arn)
+
+        while (
+            stack_status := self.cf.describe_stacks(StackName=stack_name)["Stacks"][0]["StackStatus"]
+        ) != "DELETE_FAILED":
```
Secretions, line 610: Might be worth checking for `in ["DELETE_FAILED", "DELETE_COMPLETED"]`. If either the stack was already deleted and we made it here, or the contents were already wiped out without the stack getting deleted (for whatever reason...) such that even the cloudformation-only IAM role can finish the delete, I think we would endless loop here.

Secretions (approved):

PR #144, opened 2023-08-30
Use tf import bocks

miguelhar:


convert/README.md
```diff
@@ -65,32 +69,26 @@ autoscaling group nodes and bastion node will be provisioned with the new key.
...
-    ./convert.py get-imports --region us-east-1 --stack-name mystack > terraform/imports.sh
-
-### Initialize and import Terraform resources
-
-Change to the `terraform/` directory, and run the following two commands:
+Then, we'll use `convert.py` to generate the Terraform import blocks to import our AWS resources:
 
-    terraform init
-    bash imports.sh
+    ./convert.py get-imports --region us-east-1 --stack-name mystack > terraform/imports.tf
```
Secretions, line 74: Although I'd be a tiny bit sad at losing the simplicity of just printing to stdout, I wonder if, now that everything is just terraform anyway, it would make sense to combine the `create-tfvars` and `get-imports` steps, and just have them write directly to files in some specified directory (that probably defaults to `terraform/` so you don't have to think about it 95% of th etime).

convert/lib/convert.py
```diff
@@ -634,19 +634,28 @@ def _get_nested_stacks(self, stack_name: str) -> list:
     def delete_stack(self):
         self.setup()
 
-        cf_tf_output = json.loads(
-            run(["terraform", "output", "-json"], cwd="cloudformation-only", capture_output=True).stdout
-        )
-        if not (cf_only_role := cf_tf_output.get("cloudformation_only_role")) or not re.match(
-            "arn:.*:iam:.*role/cloudformation-only", cf_only_role["value"]
-        ):
+        cf_only_state_file = os.path.join(os.getcwd(), "cloudformation-only", "terraform.tfstate")
```
Secretions, line 637: Dunno if there was a problem that was hit, but it seems more proper to get data out of terraform via the output than to dig into its data files (not that I really expect it to matter much either way).

miguelhar: There is something weird happening with the encoding on terraform 1.5 thats breaking the json_load. I think querying the state is safer and more deterministic

convert/README.md
miguelhar: I am of the same mind, even debated with `run the following commands to populate all the vars` but I landed on the same conclusion of just having the convert do this.

convert/lib/convert.py
Secretions: If it solves a problem, that's cool. Philosophically, the output should be an API that terraform can use between it and external things to ensure that data interchange format is predictable, as where we shouldn't assume the statefile will have any particular format. Pragmatically, it sounds from what you're saying that the opposite is true today, sooo... :)

PR #145, opened 2023-09-06
PLAT-7231: CDK Continuity Fixes

Secretions:
We were deprecating after k8s 1.25 (and still will), but some needs have come up to maintain continuity in testing, as well as tide people over until we have the multi-module terrraform + convert process released.

* NodeJS 14 and Python 3.7 lambdas are being deprecated. We won't be able to deploy them (even for testing) after November 27th.
* NodeJS 14 is perma-baked in to CDK v1's EKS lambdas
* Users upgrading manually to 1.25 weighed against managing calico manually, for the interim, is probably better to just manage calico manually
  * Calico being the blocker that made us decide to deprecate
  * Really, they should just wait, but it's an option if they choose not to.
  * I was iffy on whether to do this in isolation, but with the needs above for continued testing I decided it made sense.

Changes:

* Just remove calico from 1.25 on
  * Use an output to message how to get calico back
* Update to CDK v2 to get newer lambdas
* Set our 3.7 lambdas to newer python
* Manually set up the old master role, and switch our kubeconfig generation instruction to use it
  * The creation role no longer allows any random person to assume it


Secretions:
If anybody who sees this is wondering, I also tested converting this with the multi-module terraform converter (PR #146, commit [38c0add](https://github.com/dominodatalab/cdk-cf-eks/commit/38c0add5bb8cb4f74490fff4b6ff7598f6df2d43)), and it worked great without any changes. Not expecting any issues with compatibility, and should remain representative to test against going forward. The one meaningful difference (the creation role perms change) isn't an issue, since our terraform rewrites the perms.

miguelhar (approved):

Secretions:
Going to go ahead and merge this, since this is the cdk side of the work that needs to be done and there's no getting away from that, but after testing the platform-apps side of things it turns out we'll need some non-insignificant terraform/deployer changes (and all just to continue testing legacy stuff, bleh). We can no longer customize the asset bucket the way we used to, so we need to do something or other else here.

PR #149, opened 2023-09-28
PLAT-7326 fix var name

miguelhar:
[PLAT-7326](https://dominodatalab.atlassian.net/browse/PLAT-7326)

Fixes var name to match instructions.


ddl-dclegg (approved):
While I think it's still a bit confusing to have "stack" all throughout the resources and deploy_id outside, I can't see a templated reference to deploy_id, so this should work.

